PR #1, opened 2015-03-13
Cleanup

jeroenooms:


hadley:
BTW I removed xls2csv because we don't need it, but it's useful for code examples

PR #2, opened 2015-03-16
Update libxls

jeroenooms:


hadley:
Would you mind rebasing/merging please? I updated libxls at the same time as you :/

PR #20, opened 2015-03-20
Make extensions case-insensitive and support .xlsm

### LINKED ISSUE #19: Won't read non-lowercase .XLSX or .xlsm files
Looks like the code for identifying the file type is being case-sensitive with extensions, so it will open `.xlsx` files but not `.XLSX`. 

    > read_excel("epoch-1900.XLSX")
    Error: Don't know how to parse extension XLSX

Since the extension-based file type detection on platforms Excel runs on is typically case-insensitive, it seems like `read_excel` should be too, with respect to detecting file formats.

It won't read `.xlsm` files, either. They're the same format as `.xlsx`, just with macros enabled.

    > read_excel("with-macro.xlsm")
    Error: Don't know how to parse extension xlsm

apjanke:
Fixes #19.
Make extension parsing case-insensitive.
Read .xlsm files as .xlsx.

### DIFF
--- a/R/read_excel.R
+++ b/R/read_excel.R
@@ -22,10 +22,17 @@ read_excel <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
   path <- check_file(path)
   ext <- tools::file_ext(path)
 
-  switch(ext,
+  format <- switch(tolower(ext),
+    xls = "xls",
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

Issue #39, opened 2015-03-27
Cells with partial colouring read the first colour chunk?


jmarshallnz:
I have an excel sheet with a cell containing "H2395" where the "H" is in red, and the "2395" is in black.

On reading in, this reads as "H" and drops the number.  This file for example:

https://dl.dropboxusercontent.com/u/41868071/Results%20Plate%20SF325.xlsx

(On a sidenote, is it possible to retrieve the colour of a cell entry at all?  I have a bunch of sheets where (unfortunately) someone decided to store data based on the colour of the text in the cell...  Just "colour of first text" would do the trick.)

jmarshallnz:
See #41 for a fix.

PR #41, opened 2015-03-27
Combine all rich-text formatted substrings when reading the string table (#39)

jmarshallnz:
Ensures all separately formatted substrings in a cell are concatenated together as the CT_RElt cell type specifies.

Fixes issue with the spreadsheet here for example:

https://dl.dropboxusercontent.com/u/41868071/Results%20Plate%20SF325.xlsx

I suspect that strings specified inline (rather than in the stringtable) can also have similar richtext formatting, but haven't seen any such sheets.


jmarshallnz:
Done a bit more searching.  By the looks `<si>` can also include multiple `<r>` elements, as well as multiple `<rPh>` elements, each of which must contain a single `<t>` element.  Further, `<si>` can also contain a `<t>` element as well as `<r>` and `<rPh>` elements according to the spec (CT_Rst, pg 3893).

This applies both to the string table as well as single string cells as well.

So I guess one could iterate over the `<r>` elements and then append the `<t>` if non-null?  Not sure how to handle the phonetic stuff though - it's not strictly part of the displayed text I guess.

Let me know if you want me to update the PR in anyway.

jmarshallnz:
Update (apologies for the spam - last one before Monday when I'm on win32...)

Checked to see if combined `<r>` and `<t>` elements are processed (according to the spec they are).  OSX's xlsx support ignores the `<r>` tags if a `<t>` tag is present.  Doesn't matter the order.

Example files are here (cell to look at is column 1 row 11)

https://www.dropbox.com/sh/xvlkuogxo852rfb/AAAYC_i4rC3J0pIhnYc7XKXYa?dl=0

This suggests code should be:
```
       rapidxml::xml_node<>* t = string->first_node("t");
       if (t != NULL) {
         // t elements trump any r elements present
         out = std::string(t->value());
       } else {
        // iterate over all r elements
        for (rapidxml::xml_node<>* r = string->first_node("r"); r != NULL;
             r = r->next_sibling("r")) {
          // a unique t element should be present (CT_RElt type, pg 3893)
          rapidxml::xml_node<>* t = r->first_node("t");
          if (t != NULL)
            out += t->value();
       }
```
(With additional support for validating if no `<r>` elements are found if necessary).

hadley:
Looks reasonable (although I need to do a deeper read on Monday). Would you mind also adding a test?

jmarshallnz:
Sure. Will add over the weekend, and verify the test files behave as expected in Excel on Monday.

hadley:
Thanks for validating my hope that once there was clean C++ code other people might be able to understand it and fix bugs :grin:

jmarshallnz:
Ok, updated with what I think is the best way to go.  It appears Excel 2010 doesn't follow the spec, and nor does Mac OSX's Preview.  Ofcourse, they don't follow it in different ways.

I've decided to go with the spec, with the exception of a `<t>` element always being reported first, regardless of the order in the .xslx.  Excel2010 considers a `<t>` element following an `<r>` element to be invalid, so this will work fine with any Excel2010 produced .xlsx (which doesn't seem to produce mixed t/r elements anyway).

Let me know if you want any changes - most happy to do so.

dpritchard:
FWIW The fixes proposed here also resolve an issue with mixed strings containing italics.

e.g. *Macrocystis pyrifera* adult 

I realise this doesn't contribute very much concrete help, but thanks anyway!

hadley:
Thanks!

PR #43, opened 2015-04-02
Empty first column results in invalid data frame object (nrow(df) != length(df[,1]))

jmarshallnz:
This is due to the way R stores data frames by the looks - nrow() checks the row.names attribute, so the actual columns may have a different length if you try hard enough :)

Failed as Rf_length() was run on the first column which was blank.  As we remove the blank columns anyway, just run it a bit later on as we form up the data.frame.

Test included, dropped into missing.values file - let me know if you want a separate .R


hadley:
Good catch - thanks!

PR #44, opened 2015-04-02
Fixes for CMD check

jeroenooms:
I think this is the most elegant (least ugly) way of fixing this. We don't change the actual upstream code but instead mask the illegal functions with the corresponding legal alternatives.


hadley:
Ah nice!

PR #45, opened 2015-04-02
Dont drop empty named columns

jmarshallnz:
This is proof of concept for keeping empty, named, columns in the data frame passed back to R.

The first commit contains the test case.  Here the second column named 'y' has no entries.  read_excel will drop y from the resulting data.frame, whereas I think most users would expect the column to remain (it's not completely devoid of data after all as it has a name).

EDIT: Better solution implemented to alter colTypes() to take into account the names. Will need similar changes for .xls as well assuming this is the way forward.


jmarshallnz:
Updated to be a bit more sane by just altering colTypes() to take into account the names (i.e. if name is in the sheet, don't mark the column as `CELL_BLANK`).

Will need similar code added for .xls as well, assuming this is the way forward.

hadley:
Hmmmmm. If we go this root I think it should be an option, and I think it would be easier to just produce something for a column of type cell blank (e.g. a `LogicalVector` full of NAs). Does that make sense?

hadley:
(i.e. just modify `makeCol()`

hadley:
And it would be good to pair `skip_empty_rows` and `skip_empty_cols` as arguments (although they're a bit long as names)

jmarshallnz:
That is another option, sure, along with the checking of a boolean in `colDataframe()`.

I'm not sure if there should be a distinction between columns that are completely empty (no header at all) and those that have a header and are otherwise empty.  For the sheets I'm dealing with at the moment for example, I don't want the former (no data there) but do want the latter (the absence of data in a column that should have data is informative).

hadley:
Ah, got it, that makes sense. Want to have a go at the xls code too?

jmarshallnz:
Sure - just pushed up what I have (untested).

jmarshallnz:
Ok, updated with code + text for .xls as well as .xlsx.

R/RcppExports.R
```diff
@@ -17,14 +17,14 @@ xls_sheets <- function(path) {
...
 
-xls_col_types <- function(path, na, i = 0L, nskip = 0L, n = 100L) {
-    .Call('readxl_xls_col_types', PACKAGE = 'readxl', path, na, i, nskip, n)
-}
-
 xls_col_names <- function(path, i = 0L, nskip = 0L) {
     .Call('readxl_xls_col_names', PACKAGE = 'readxl', path, i, nskip)
 }
 
+xls_col_types <- function(path, na, sheet = 0L, nskip = 0L, n = 100L, has_col_names = FALSE) {
```
hadley: What version of Rcpp do you have? I think this spurious diff is caused by an older version

R/read_excel.R
```diff
@@ -40,19 +40,19 @@ read_xls <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
...
   }
 
   if (is.null(col_types)) {
-    col_types <- xls_col_types(path, sheet, na = na, nskip = skip)
+    col_types <- xls_col_types(path, sheet, na = na, nskip = skip, has_col_names = has_col_names)
   }
 
   xls_cols(path, sheet, col_names = col_names, col_types = col_types, na = na,
-    nskip = skip)
+    nskip = skip + ifelse(has_col_names, 1, 0))
```
hadley: That's just a long way of writing `skip + has_col_names` ;)

R/RcppExports.R
hadley: Oh maybe it's because the order of the functions change in the .cpp file?

src/XlsWorkSheet.cpp
```diff
@@ -7,23 +7,32 @@
...
   XlsWorkBook wb = XlsWorkBook(path);
-  std::vector<CellType> types = wb.sheet(i).colTypes(na, nskip, n);
+  return wb.sheet(i).colNames(nskip);
+}
+
+// [[Rcpp::export]]
+CharacterVector xls_col_types(std::string path, std::string na, int sheet = 0,
+                              int nskip = 0, int n = 100, bool has_col_names = false) {
+  XlsWorkBook wb = XlsWorkBook(path);
+  std::vector<CellType> types = wb.sheet(sheet).colTypes(na, nskip + (has_col_names ? 1 : 0), n);
```
hadley: Again, you can just do `nskip + has_col_names`

src/XlsWorkSheet.cpp
```diff
@@ -7,23 +7,32 @@
...
     out[i] = cellTypeDesc(types[i]);
   }
 
-  return out;
-}
+  if (has_col_names) {
+    // blank columns with a name aren't blank
+    CharacterVector names = xls_col_names(path, sheet, nskip);
+    for (size_t i = 0; i < types.size(); ++i) {
+      if (types[i] == CELL_BLANK && names[i].size() > 0)
```
hadley: Maybe `names[i] != ""` would be clearer?

R/RcppExports.R
jmarshallnz: Yup, that's the reason.

src/XlsWorkSheet.cpp
jmarshallnz: Agreed - I wanted names[i].empty() but Rcpp doesn't support it (PR to them as well).

size() is strlen() so doing a != will be faster at this point (not that the strings are long anyway) as well as being clearer.

jmarshallnz:
Updated - thanks for the review :)

hadley:
Awesome - thanks!

Issue #48, opened 2015-04-02
Warning "unkown type 'inlineStr' leads to NA column with NA header


FvD:
In a larger excel sheet there is one tab that results in an additional column with NA as a header. A function like dplyr::distinct will fail on this column later. A warning for each row in the excel sheet is generated:
````
Warning messages:
1: In read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  ... :
  [3, 4]: unknown type 'inlineStr'
2: In read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  ... :
  [4, 4]: unknown type 'inlineStr'
3: In read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  ... :
````

It may have something to do with working in both Microsoft Excel and LibreOffice with the same document. There are references to similar issues in the [python-excel](https://github.com/python-excel/xlrd/issues/81) project, and a description of the conflicting xml [here](http://our.componentone.com/groups/topic/read-generated-excel-with-inlinestr-cell-type/). The [test file](http://our.componentone.com/wp-content/uploads/2011/12/TestExcel.xlsx) in that last link will repeat the issue in readxl.

PR #52, opened 2015-04-08
Refactor string extraction to handle richtext inside inlineStr

jmarshallnz:
As noted in #48, I've pulled out the string parsing to a separate function, and added a bool flag (not particularly pretty) for whether the text is missing or not so it can also be used for inlineStr parsing.

Test added by modifying inlineStr.xlsx to produce a new file inlineStr2.xlsx, which loads successfully in Excel as well as readxl.


src/XlsxCell.h
```diff
@@ -29,6 +29,47 @@ inline std::pair<int, int> parseRef(const char* ref) {
   return std::make_pair(row - 1, col - 1); // zero indexed
 }
 
+// Parser for <si> and <is> inlineStr tags CT_Rst [p3893]
+inline std::string parseString(rapidxml::xml_node<>* string, bool *missing = NULL) {
```
hadley: I think it's customary to do this the other way aroud - `parseString()` should return a bool indicating success/failure and have an output param which is the string. 

The coding convention I follow is that all pointers should have a p prefix, so I think a good name here would be `pOut`

src/XlsxCell.h
```diff
@@ -29,6 +29,47 @@ inline std::pair<int, int> parseRef(const char* ref) {
...
+    //       Any other combination of r and t elements is considered invalid.
+    //
+    //       MacOSX preview, however, considers only the t element of mixed
+    //       r/t elements.
+    //
+    //       The spec seems to allow a single t and zero or more r elements
+    //       to coexist.
+    //
+    out = std::string(t->value());
+    found = true;
```
hadley: Does it need a return here? That would agree with the comment "t elements trump any r elements"

jmarshallnz: I'll reword the comment to make clear that we follow what Excel allows (t first with as many r's as you like after).

jmarshallnz:
Updated.

src/XlsxCell.h
```diff
@@ -29,6 +29,49 @@ inline std::pair<int, int> parseRef(const char* ref) {
   return std::make_pair(row - 1, col - 1); // zero indexed
 }
 
+// Parser for <si> and <is> inlineStr tags CT_Rst [p3893]
+// returns true if a string is found, false if missing.
+inline bool parseString(rapidxml::xml_node<>* string, std::string &out) {
```
hadley: Can you rename `out` to `pOut` please?

hadley: Oh, and make it a `std::string* out` - that makes it more clear that the input is being modified

jmarshallnz: Happy to do so. I generally prefer references in cases like this where you're guaranteed that the (out/in)put exists and is writable (passing NULL to the function is clearly an error), and pointers where it's optional, as any pointer adds burden of existence checks.

Should I return false on NULL (confusing the return code), or assert (my preference, not sure how useful it is within an R library), or ignore?

I guess I should also make `string` const, and check that for NULL as while a NULL node is currently programmer error, it may not be in future, and doesn't confuse the return code (missing node -> missing value).

jjallaire: Our code style doesn't use references for out parameters because it's not self-documenting at the point of *passing* and *using* the pointer/reference. You are passing a variable to a function which might mutate it, and doing this in the following style:

```cpp
foo(&bar);
```

Makes this extremely clear to the reader. This policy combined with the policy that NULL pointers are never allowed and smart pointers are used anytime there are non-trivial ownership semantics achieves the safety you are looking for (i.e. by convention you never have to check a raw pointer for NULL and never have to delete it -- raw pointers are always valid and always owned by someone else).

These idioms were adopted from the Google C++ Style Guide, which specify that no non-const references should be used as function arguments,  see https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Reference_Arguments and this presentation which goes into more depth on the non-const reference issue at the end: http://goo.gl/qXprar

jmarshallnz: Thanks - appreciate the explanation. Will update shortly.

jmarshallnz:
Updated

hadley:
Thanks!

(Hmmm, just realised that I probably shouldn't accepted this on to master because I just submitted to CRAN. Will have to remember to tag correctly)

Issue #51, opened 2015-04-07
New line in cell can create _x000D_


janzzon:
Some types of new lines in excel-cell looks ok in excel, but adds
``` _x000D_ ``` when imported with read_excel.
Example in my dropbox, file "New line errors.xlsx"
https://www.dropbox.com/sh/8mzk6z12f99ye4w/AABe-C7fxCzm1lomNyge1pw_a?dl=0

**Comment by hadley** [comment]
How did you create the file? If I resave it, it seems to load in ok.

The xml literally looks like this:

```xml
<sst xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" count="3" uniqueCount="3">
  <si>
    <t>column_name</t>
  </si>
  <si>
    <t>first line of text _x000D_
second line of text _x000D_
i don&apos;t know what type of line feed I&apos;ve used _x000D_
last line</t>
  </si>
  <si>
    <t>Will show &quot;_x005F_x000D_&quot; in every new line in cell above</t>
  </si>
</sst>
```

@jmarshallnz any thoughts? @evanmiller have you seen this before?

(This seems related: http://stackoverflow.com/questions/14944456/datatable-writexml-how-to-preserve-spaces-in-element-names)

jmarshallnz:
Here's another example of this.  Seems Excel is using a weird way of storing hex entities?

jmarshallnz:
And another from apache, which refers to this being in the OOXML spec (but no reference unfortunately)

https://bz.apache.org/bugzilla/show_bug.cgi?id=49653

hadley:
@jmarshallnz plus I'm pretty sure you _can_ represent a carriage return in XML (since it's a unicode character)

PR #55, opened 2015-04-10
Fix for .xlsx containing 'unicode character expressions' such as _x000D_.

jmarshallnz:
This closes out #51.  It appears Excel (assuming that's what produced the spreadsheet the provided) doesn't obey it's own advice, as 0xD is perfectly valid in XML v1.0.  In any case, this occurs, so should be supported.

As far as I could tell (with an admittedly brief look) the regex stuff in R is not available on the C side, and as this was an easy thing to look for at any rate, the iterative code isn't too bad given we stay in C-land.

Note the use of the GraphicsDevice.h function (clearly someone else needed to do this at some point!) which at this stage is just exported directly until the code is finalised, but should really be pulled in via the include, which has the same definition.  Using Riconv directly is ugly, thus the existence of this function!

Another alternate would be to include the 15 lines or so required to convert unicode (i.e. utf32) codepoints to utf8 which is easy enough (and would be faster as no need to construct the Riconv object or perform the buffer manipulation).

Finally, another alternate would be to jump to R land and do something like `eval(parse("'\\u000D'"))` after first constructing that string via gsub to get R to parse each chunk directly. (Speed probably isn't an issue here).


hadley:
I worry a little bit about including this unconditionally. I don't think it's every version of excel that uses this strange escaping - if I open and resave the problematic file, the line breaks are encoded correctly. So maybe just wrap this in a function that the user can call?

jmarshallnz:
I guess there is a small chance that this would decode otherwise valid data as something else, but that is also what Excel itself does on loading the test sheet.

If a separate function, I guess it would be a function to process a data frame that processes all character/factor columns?

jmarshallnz:
I've loaded the sheet in a bunch of other applications, and they all respond the same way (i.e. by decoding the encoded characters).

I think if you add `_x####_` to any excel spreadsheet, it'll double-escape the string on save (by escaping the first underscore), which suggests the decoding should always be performed.  I'll confirm this Monday

hadley:
Ok, in that case, I'd be happy to accept this change (although I wouldn't mind a quick benchmark to see if the performance impact is big - I guess it won't be a problem in most cases since it will only be applied once to the string table)

jmarshallnz:
Sure - will do so in a couple days (teaching campus course today).

Do you want me to factor out the stringParse + unescape functions to a separate file (XlsxString.h ?)

hadley:
Eh, it's up to you - I don't have strong feelings either way

jmarshallnz:
Updated and rebased. I've benchmarked using a highly unrealistic example of two 5000x1000 spreadsheets filled with random 4 letter strings (around 500k of them) or with random 10 letter strings (500k of them).  Can't really detect a difference in load time (i.e. more variability within than between) from the R side.

Benchmarking the time taken to parse the string table on the C++ side shows a reduction in speed from 63ms to 92ms or 112ms to for the above examples. Total load times into R are between 5 and 7 seconds, so the slow down is less than 1% of the total.

Given the above are worst-case scenarios (most .xlsx likely have numbers in them as well) I reckon we can discount further optimisation :)

hadley:
Awesome - thanks! 

Now that the first version is up on CRAN, can you please add a bullet to `NEWS.md` that briefly describes the change?

jmarshallnz:
Done - not completely sure of the format you use :)

hadley:
Thanks!

PR #57, opened 2015-04-10
add repo for cyclic dep check, add notifications

bquast:


hadley:
Do you need those notifications? I think they're just the defaults.

bquast:
I suppose not, I added them because they are in the devtools template, I believe Craig Citro put them there. The repo is helpful though, having no notes is nice and clean

bquast:
I've also created a pull request with the repo added in the devtools template

bquast:
I've removed the notification lines from request

hadley:
Thanks!

PR #77, opened 2015-04-16
added url capability to read_excel

jentjr:
Here is an example data set to try.

`idem_url <- "http://www.in.gov/idem/files/groundwater_gwmn_data_2008.xls"`

`gw <- read_excel(idem_url)`


R/read_excel.R
```diff
@@ -26,13 +26,26 @@ NULL
 read_excel <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
                        na = "", skip = 0) {
 
-  path <- check_file(path)
-  ext <- tolower(tools::file_ext(path))
+  if (substring(path, 1, 7) == "http://" ||
```
PedramNavid: Might also want to add https:// to path search.

pssguy:
Why would I get this error. I'm on CRAN version 1.1
```
idem_url <- "http://www.in.gov/idem/files/groundwater_gwmn_data_2008.xls"

gw <- read_excel(idem_url)

Error: 'http://www.in.gov/idem/files/groundwater_gwmn_data_2008.xls' does not exist in current working directory ('C:/Users/.....')

```

jennybc:
@pssguy If you're running the CRAN version, as opposed to the version in this pull request, `read_excel()` has no understanding of URL as input. So it's looking for an actual file by that name, which presumably doesn't exist.

pssguy:
@jennybc Thanks for reply. So this pull request never got merged? Any idea why not. This is not a new issue

jennybc:
Nothing has been merged for a while, but this is my assignment now 🙂.

I am still triaging and grouping issues, but then things will happen.

pssguy:
Good to hear. Do you know of an alternative option of directly importing?

I would like to do the equivalent of this

df <- read_excel("https://github.com/tts/villisika/blob/master/SpeciesObservations.xlsx",, sheet = "SLW Data")

Obviously no problem obtaining data and loading from local file but am looking to have it as line in notebook code available on web

jennybc:
> Do you know of an alternative option of directly importing?

At the moment, no, I am not aware of anything you can do with readxl here other than download the file, then read. I don't see why you couldn't do that in a notebook.

hadley:
We're going to attack this problem a little more broadly. Thanks for your work!

jebyrnes:
Curious - I just installed the github version (1.3.1.9000), yet still this appears to be having issues. 

```
read_excel("https://fragilestatesindex.org/wp-content/uploads/2019/04/fsi-2019.xlsx")
```

produces the following 

```
Error: `path` does not exist: ‘https://fragilestatesindex.org/wp-content/uploads/2019/04/fsi-2019.xlsx’
```

Yet this file downloads perfectly fine from, say, a browser.

N.B. I found this while tinkering about for this week's #tidytuesday so the issue might pop up again.

jennybc:
We have not yet implemented any sort of "read directly from URL" in readxl.

PR #83, opened 2015-04-18
test xls/x saved as ods files

mikerspencer:
Hi Hadley, as discussed on twitter; here are the test files saved as .ods using libreoffice v4.2.7.2 on Ubuntu 14.04.

Let me know if I can help with anything.


hadley:
@jennybc can you please start adding these to the tests

jennybc:
I don't have the Twitter discussion for context ...

Is this the idea?

Test that we get the same R object when we `read_excel()` on the same spreadsheet, when it has been written by ... Excel into xlsx, Excel into xls, LibreOffice, etc.?

mikerspencer:
Yes, I'm hoping we can use `read_excel` (`read_ods`?) to read open document spreadsheets into R. Although, perhaps the idea has evolved since I suggested it!

jennybc:
> Yes, I'm hoping we can use read_excel (read_ods?) to read open document spreadsheets into R.

Oh that's a big job!

@hadley I propose to close this pull request and open + link to an issue about reading `.ods` files. If that glorious day comes, we would then bring `.ods` versions of test and example sheets into readxl. Until then, I don't think it makes sense to have them here.

I thought this might be about reading `.xlsx` written by LibreOffice, but it seems it's not.

hadley:
Oh I thought `.ods` files were very close to `.xlsx` and so this would be an easy job. I think you're right that it would be better to start with an issue.

jennybc:
`.ods` are zip archives of xml, but I think that is where the similarity ends. It's doable but it would definitely be adding a 3rd format. Maybe there would be some overlap with the `.xlsx` code?

In the meantime @mikerspencer, I assume you know about [readODS](https://github.com/chainsawriot/readODS#readme).

mikerspencer:
Thanks Jenny, yes I've been finding other ways around the issue!

Sent from my phone, apologies for typos.

website: mikerspencer.com

PR #98, opened 2015-05-08
WIP: Equality of autogenerated colnames (xls vs. xlsx)

krlmlr:
With `col_names = FALSE`, .xls colnames start with X0 and .xlsx colnames start with X1. Shall column names always begin with X1?

So far, this pull request only contains a failing test.

Test output:

```
1. Failure (at test-colnames.R#7): auto-generated column names are the same for xls and xlsx files 
colnames(xls) not equal to colnames(xlsx)
2 string mismatches:
x[1]: "X0"
y[1]: "X1"

x[2]: "X1"
y[2]: "X2"
```

(Funny: Another real-life Excel file shows the same behavior **the other way round**: [xls](https://dl.dropboxusercontent.com/u/5686591/be-b-00.04-rgs-14.xls) [xlsx](https://dl.dropboxusercontent.com/u/5686591/be-b-00.04-rgs-14.xlx))

### DIFF
--- a/tests/testthat/test-colnames.R
+++ b/tests/testthat/test-colnames.R
@@ -1,6 +1,6 @@
 context("Colnames")
 
-test_that("dates respsect worksheet date sheeting", {
+test_that("auto-generated column names are the same for xls and xlsx files", {
   xls <- read_xls("missing-values.xls", col_names = FALSE)
   xlsx <- read_xlsx("missing-values.xlsx", col_names = FALSE)
 
--- a/tests/testthat/test-colnames.R
+++ b/tests/testthat/test-colnames.R
@@ -1,0 +1,8 @@
+context("Colnames")
+
+test_that("dates respsect worksheet date sheeting", {
+  xls <- read_xls("missing-values.xls", col_names = FALSE)
+  xlsx <- read_xlsx("missing-values.xlsx", col_names = FALSE)
+
+  expect_equal(colnames(xls), colnames(xlsx))
+})


hadley:
Oooh yes, that's probably an off-by-one error in my C++ code

zeehio:
(Just in case you are interested) I fixed this issue in #165, because my test was failing due to this bug.

Courvoisier13:
Hi,
is there a reason why you guys added a double underscore "__" for the column names? X__1 instead of X1?
thanks

krlmlr:
I don't know, could you please start a new issue with a reproducible example?

Courvoisier13:
I am not sure it's an issue. see https://github.com/tidyverse/readxl/releases it says:
Now, the default column names are X__1, X__2, X__3. I don't understand why there is a double underscore.

krlmlr:
Neither do I, but discussions in closed issues rarely seem to be helpful. That's why I suggested to open a new issue or perhaps ask elsewhere. (Maybe there's a relevant item in the NEWS?)

zeehio:
The column names now are X1, X2... as expected. This looks like an error in the NEWS file introduced at https://github.com/tidyverse/readxl/commit/a41f61b547391781b2931c90cd9b6aa83537cda3 , so don't worry about it (or submit a pull request fixing it if it bothers you a lot).

Courvoisier13:
No, I am getting X__1, X__2..... huh, looks like we get different results. will open an issue and post examples.

PR #100, opened 2015-05-17
Export read_xls(), read_xlsx(): Closes #85

### LINKED ISSUE #85: Add option to specify file format to support Shiny fileInput()
When a file is loaded through a fileInput() widget in a Shiny app, it appears that read_excel scans the file name to determine the file type (at least when input$file[,4] - the file path) is used.  Shiny uploads an extension-less temporary file so read_excel can't open the file.

jirkalewandowski:


hadley:
I'm not sure if this is the way I want to solve this problem. I think it would be more consistent with other packages (e.g. haven) to expose xls_sheets & xlsx_sheets etc.

jennybc:
I have some hope this will get fixed on the Shiny side (preservation of file extension and/or name).

But even so, there are several requests to handle more general forms of input (`rawConnection` #192, URL #183, compressed files #115), so it seems reasonable to expose `read_xls()` and `read_xlsx()` in the long run.

I'm editing the title of this PR to better reflect the content.

tests/testthat/test-read-excel.R
```diff
@@ -13,3 +13,16 @@ test_that("can read sheets with inlineStr", {
   x <- read_excel("inlineStr.xlsx")
   expect_equal(x$ID, "RQ11610")
 })
+
+test_that("can read file without ending", {
+
+  file.copy("iris-excel.xlsx", "iris-xlsx-no-ending")
+  x <- read_xlsx("iris-xlsx-no-ending")
+  expect_error(read_xls("iris-xlsx-no-ending"))
+  file.remove("iris-xlsx-no-ending")
```
jennybc: Can you do file removal using `on.exit()` instead? Here and below.

Example elsewhere: https://github.com/hadley/xml2/blob/bfcefb98b4da4f3630ac781ecb2b77b0a2b444ab/tests/testthat/test-write_xml.R#L16

tests/testthat/test-read-excel.R
```diff
@@ -13,3 +13,16 @@ test_that("can read sheets with inlineStr", {
   x <- read_excel("inlineStr.xlsx")
   expect_equal(x$ID, "RQ11610")
 })
+
+test_that("can read file without ending", {
+
+  file.copy("iris-excel.xlsx", "iris-xlsx-no-ending")
+  x <- read_xlsx("iris-xlsx-no-ending")
+  expect_error(read_xls("iris-xlsx-no-ending"))
```
jennybc: Can you include the error message here via `regex`?

tests/testthat/test-read-excel.R
```diff
@@ -13,3 +13,16 @@ test_that("can read sheets with inlineStr", {
...
+  file.remove("iris-xlsx-no-ending")
+
+  file.copy("iris-excel.xls", "iris-xls-no-ending")
+  x <- read_xls("iris-xls-no-ending")
+  file.remove("iris-xls-no-ending")
```
jennybc: To be symmetric with the above, shouldn't this check for error reading extension-free xls with `read_xlsx()`?

tests/testthat/test-read-excel.R
```diff
@@ -13,3 +13,16 @@ test_that("can read sheets with inlineStr", {
   x <- read_excel("inlineStr.xlsx")
   expect_equal(x$ID, "RQ11610")
 })
+
+test_that("can read file without ending", {
+
+  file.copy("iris-excel.xlsx", "iris-xlsx-no-ending")
+  x <- read_xlsx("iris-xlsx-no-ending")
```
jennybc: Maybe compare this to result of `read_excel("iris-excel.xlsx")`? Ditto for xls.

jennybc:
Thanks! Will you also add a bullet to NEWS?

tests/testthat/test-read-excel.R
jennybc: This is still needed. We have some (weak) convention where we try to specify all or part of the expected error message.

jirkalewandowski:
Wow, thanks for coming back to this and your feedback! I did not implement
[checking for the error message via regexp](https://github.com/hadley/readxl/pull/100#pullrequestreview-18248455), since
the error message/behavior readxl gives in these cases seems not very helpful to me:

```
> load_all()
Loading readxl
> read_xls("tests/testthat/blanks.xlsx")
Not an excel file
 Show Traceback
 
 Rerun with Debug
 Error in xls_col_names(path, sheet, nskip = skip) : 
  Failed to open tests/testthat/blanks.xlsx
> read_xlsx("tests/testthat/blanks.xls")
 Show Traceback
 
 Rerun with Debug
 Error in read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  : 
  zip file '/home/jlewando/data/github/readxl/tests/testthat/blanks.xls' cannot be opened
```

The line "Not an excel file" is printed from [src/ole.c](https://github.com/hadley/readxl/tree/master/src/ole.c#L301) and not passed up as an error.
I wonder if there is a good solution to check for the filetypes up front which is not extension based...?

Regarding the reading from URLs, rawConnection, ... I took a look at it seems like the Cpp is written
to read from paths. So I think one solution would be to expand [check_file](https://github.com/hadley/readxl/tree/master/R/utils.R#L1) to take any
of these types, and always return a path. I.e. download/write/... transform the input. Do you see a better solution?

jennybc:
> I did not implement checking for the error message via regex

Yeah, I agree the error messages aren't great for the user, but let's drop `Not an excel file` and `zip file '\\.+' cannot be opened` in there anyway. I'll try to improve that situation when I'm in the right neighbourhood.

> Regarding the reading from URLs, rawConnection...

I think the solution for this might reside outside of readxl. In any case, it's not an immediate goal. I just brought it up because it's more justification for exporting `read_xlsx()` and `read_xls()`. It will need to happen eventually anyway. Even with their current functionality, we can at least work around Shiny stripping the file extension!

PR #119, opened 2015-08-07
Improve error messages

warnes:
to provide more information.
- For read_xlx, when the requested worksheets doesn't exist.
- For read_xlx, and read_xlsx, when the number of column names or types doesn't match the file.

### DIFF
--- a/src/XlsWorkSheet.h
+++ b/src/XlsWorkSheet.h
@@ -6,6 +6,7 @@
 #include "XlsWorkBook.h"
 #include "CellType.h"
 #include "utils.h"
+#include <string>
 
 class XlsWorkSheet {
   xls::xlsWorkSheet* pWS_;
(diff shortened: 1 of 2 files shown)


src/XlsWorkSheet.h
```diff
@@ -18,9 +19,10 @@ class XlsWorkSheet {
   XlsWorkSheet(const XlsWorkBook& wb, int i) {
     offset_ = dateOffset(wb.workbook()->is1904);
 
-    if (i < 0 || i >= wb.nSheets())
-      Rcpp::stop("Invalid sheet index");
-
+    if (i < 0 || i >= wb.nSheets()) {
+      Rcpp::stop(Rcpp::sprintf<80>( "Invalid sheet index.  Should be in [0, %d]", wb.nSheets() ).c_str());
```
hadley: I'm pretty sure `stop()` already wraps `sprintf`, so you can just do: `Rcpp::stop("Invalid sheet index.  Should be in [0, %d]", wb.nSheets());` . If not, I'd recommend using tinyformat which Rcpp bundles.

Why `%d` and not `%i`?

behrica:
Very usefull pach. Without it I cannot 'guess' how many blank columns I have.

jennybc:
I'm afraid this got handled in a larger refactoring of column specification and naming (#261), so I'm closing this. Thanks!

PR #130, opened 2015-09-20
Added capability to use one column type as type for all existing columns.

rdoering:
It also tells the user how how many colums exists at the excel file, if it fails with "Error: Need one name and type for each column".


hadley:
Thanks for this, but I think we'll solve a little more generally.

PR #56, opened 2015-04-10
Support multiple na values

jmarshallnz:
Takes care of #13, and fixes a bug where NA = "1" didn't work for .xls sheets.

PR #136, opened 2015-10-12
read_excel now accepts a vector of NA types

MichaelChirico:
Can be tested on this test Excel file:

https://drive.google.com/file/d/0B1i6iIc4UbO1VUV6TWQ0RjlFT1U/view?usp=sharing

Using this code:

```
read_excel("test_mult_na.xlsx",sheet=1,col_names=T,
           col_types=c("text","numeric","numeric"),na=c("","-","NA"))
```

I get this output, as expected:

```
  var1     var2 var3
1    a    1.300    1
2    b       NA    2
3    c   51.500    3
4 <NA>    9.900    4
5    e 8548.400    5
6    f  848.560    6
7    g   32.498   NA
```

### DIFF
--- a/src/CellType.h
+++ b/src/CellType.h
@@ -48,13 +48,13 @@ inline std::string cellTypeDesc(CellType type) {
 
 inline CellType cellType(xls::st_cell::st_cell_data cell, xls::st_xf* styles,
                          const std::set<int>& customDateFormats,
-                         std::string na = "") {
+                         std::vector<std::string> na = {""}) {
   // Find codes in [MS-XLS] S2.3.2 (p175).
   // See xls_addCell for those used for cells
   switch(cell.id) {
   case 253: // LabelSst
   case 516: // Label
-    return (na.compare((char*) cell.str) == 0) ? CELL_BLANK : CELL_TEXT;
+    return (std::any_of(na.cbegin(), na.cend(), [&](std::string i){ return i.compare((char*) cell.str) == 0; })) ? CELL_BLANK : CELL_TEXT;
     break;
 
   case 6:    // formula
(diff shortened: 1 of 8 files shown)


MichaelChirico:
Not sure why Travis / `R CMD check --as-cran` failed, as the latter works on my local machine.

src/CellType.h
```diff
@@ -48,13 +48,13 @@ inline std::string cellTypeDesc(CellType type) {
 
 inline CellType cellType(xls::st_cell::st_cell_data cell, xls::st_xf* styles,
                          const std::set<int>& customDateFormats,
-                         std::string na = "") {
+                         std::vector<std::string> na = {""}) {
```
hadley: That needs to be a `const std::vector<std::string>&` otherwise the complete vector gets copied each time

src/CellType.h
```diff
@@ -48,13 +48,13 @@ inline std::string cellTypeDesc(CellType type) {
...
   // Find codes in [MS-XLS] S2.3.2 (p175).
   // See xls_addCell for those used for cells
   switch(cell.id) {
   case 253: // LabelSst
   case 516: // Label
-    return (na.compare((char*) cell.str) == 0) ? CELL_BLANK : CELL_TEXT;
+    return (std::any_of(na.cbegin(), na.cend(), [&](std::string i){ return i.compare((char*) cell.str) == 0; })) ? CELL_BLANK : CELL_TEXT;
```
hadley: We can't rely on C++11 yet. Could you please extract out into a separate helper function that just uses iteration?

src/CellType.h
hadley: Is that default value a C++11 extension?

MichaelChirico: @hadley Thanks for the tip! This is honestly my first piece of C++ code so I can't say I have the slightest idea which bits are 11-specific or not.

Before repairing this pull, I noticed after submitting this request that there's an [outstanding pull request](https://github.com/hadley/readxl/pull/56) that seems to accomplish the same thing and from what I see was just about ready to be merged, was there something wrong with that code that I missed?

Don't want to duplicate efforts here.

hadley:
Closed as duplicate of #56

Issue #131, opened 2015-09-25
Feature request: add parameter for number of rows to read to impute column type


markriseley:
read_excel works fast in part because it imputes col types on first x rows only (is it 100?). The flipside is it doesn't handle sparse columns well (see this SO post: http://stackoverflow.com/questions/6099243/read-an-excel-file-directly-from-a-r-script).

It would be helpful to expose this parameter to the user eg 'check_len = 100L', perhaps with '-1' to represent check all rows before assigning column types. [should also be applicable to readr::type_convert]

Supplying coltypes in advance may not be practical in some scenarios (eg when parsing file uploads in a shiny application rather than in interactive use).

My current workaround is to use unexported functions eg xlsx_dim to get number of columns (due to #114,#127), then col_types = rep("text",ncol), then type.convert(as.is = T).

Example use case: reading in market research data, where multiple response questions are often coded as either blank or some text representing the response code rather than T/F and can often contain large tracts of blank rows.

elastadmin:
Having the same issue. In my case, the first cell of data in a given column is blank and the actual data only starts at row 150. read_excel assumes it is numeric in data type so when it hit character data, it forces them all to NA. Might be better to default to character datatype if type cannot be inferred. Better to convert a column to the appropriate datatype later than losing data. Pre-specifying datatypes won't work all the time as I don't control the file format sent to me.

hangtime79:
+1 for this feature request:
I followed to this forum from this StackOverflow answer: http://stackoverflow.com/questions/31633891/specifying-column-types-when-importing-xlsx-data-to-r-with-package-readxl 

In a case where there is data sparsity in a column the ability to specify the number of rows in order to determine the column type would be ideal. Based on the StackOverflow answer it appears that ReadXL may be able to get to that easier than others. This would be a great feature to have to an already outstanding package. Thanks for the hardwork!

garthtarr:
+1 for this feature request, `check_len` would be an exceptionally useful parameter!

schelhorn:
+1, urgently needed.

schelhorn:
@hadley, I know you're busy but is there any chance this could be implemented earlier rather than later? Thanks a bunch.

tklebel:
I tried an implementation and created a PR.

Until hadley merges it you can test it out with `devtools::install_github("tklebel/readxl")`.

schelhorn:
Spitze, danke!

tklebel:
[...]

Issue #112, opened 2015-06-18
Column type is determined by first 100 rows.  Please document.


kwstat:
In this file: https://github.com/hadley/readxl/blob/master/src/XlsxWorkSheet.h

Is this line of code:
```
std::vector<CellType> colTypes(const std::string& na, int nskip = 0, int n_max = 100, bool has_col_names = false) {
```
That appears to use the first 100 lines of data to determine the column type.  Please document this in the appropriate Rd file.  (At least, I didn't see it).  The reprex below shows what can happen if there are 100 blanks followed by data values.

```
library(openxlsx)
library(readxl)

maxrow=100
examp <- data.frame(a=1:maxrow, b=c(rep(NA,maxrow-1), "R1"))
openxlsx::write.xlsx(examp, file = "examp_100.xlsx")
read_excel("examp_100.xlsx")

maxrow=101
examp <- data.frame(a=1:maxrow, b=c(rep(NA,maxrow-1), "R1"))
openxlsx::write.xlsx(examp, file = "examp_101.xlsx")
read_excel("examp_101.xlsx")
```
The second example gives the following warning and does not read the "R1" value in row 101
```
[rows 1:98 omitted]
99   99 NA
100 100 NA
101 101 NA
Warning message:
In read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  :
  [102, 2]: expecting numeric: got 'R1'
```

jbeaulie:
This has caused me some grief.  I have a file with >200,000 rows.  The first ~190,00, rows are blank, but the remaining are populated with text.  read_excel reads the entire column as NA.

tklebel:
Some discussion happened in #131, I think we could close this issue here.

There is an open pull request to read 1000 rows by default. #143 

I forked `readxl` and implemented a parameter to choose the number of rows to guess the `col_type`. There is an open PR for that too #138, but it will probably take some time until Hadley works on `readxl` again.

@jbeaulie, until then you could try out the forked version:

```R
devtools::install_github("tklebel/readxl")
read_excel("file.xlsx", n = 200)
```

kwstat:
The self-contained example at the top of this thread might still be useful.

Issues #131, #138, #143 go beyond my original request to modify the documentation and propose changes to allow more rows to determine the data type for a column.  I'll close this issue.

PR #138, opened 2015-10-17
Add param for number of rows used when guessing col_types. Implements #131

tklebel:
I gave #131 a shot. 
- Re-Rccp
- new param `n` including documentation, which passes `n` for `.xls` and `n_max` for `.xlsx` up to `read_excel`

``` r
# read 500 rows to determine col_types
read_excel("file.xlsx", n = 500)
```
- Updated tests:
  - Changed `types.xlsx`.
    - The first column has now 102 rows, all numeric.
    - Added new column `string_at_102` which contains `"ab"` at row 102.
  - added tests for correct `column_types`, `expect_warning` for `n = 100` and `expect_silent` for `n = 102`


hangtime79:
Thomas, 
I was able to download, compile, and test your code. I tested with a spreadsheet that has been giving me issues with a column with mixed data types and your code appears to be working.

image


The response column included characters but they were not found until you got to row ~6000. Great work! Thank you.

Question is n_max just the same parameter for the number of rows to test against xlsx vs n for xls or is it something else?

tklebel:
@hangtime79 the naming of the parameters internally is not very clear right now. Indeed, at some point `n_max` is used for .xlsx and `n` for .xls. If you dig deeper, they both point to `n_max` in the end.

For you it doesn't matter, since you would be using `read_excel("file.xls/x", n = 6500)` for both cases. Please tell me if that is how you used it, since that is how it was intended. Maybe my message above was a bit confusing.

@hadley I tried to stick with the names which I found in your code. Ultimately I think you should decide, how the params are named and where to make the switch from `n_max` to `n` (or any other name you would prefer for this param).

bhive01:
I think I prefer to have the option to define how many rows before guessing the type to just setting a value, with some sort of sensible value. @hadley recently updated readr to do 1000 rows before guessing, and I just did a pull to align readxl with this value. 
https://github.com/hadley/readxl/pull/143

Like I said, it'd be better to have a reasonable default (1000?), but allow user definition like this pull.

hangtime79:
@bhive01 I agree a 1000 rows is much more reasonable as per #143 but in the case of large sheets with sparse data, this particular option is incredibly valuable and doesn't impact performance of the larger population. In this case, user definition is the key. What @tklebel was able to do makes life much easier as it means you don't have to devolve into setting the data types of 80 columns in the case of some of my sheets. That's more coding on the part of the project and more on the part of the individual user.

bhive01:
I'm with you ( @hangtime79 ) and @tklebel  then. 1000 row default like `readr` and allowing user definable max at the expense of increased computation time for huge sheets. I prefer `nmax` for a variable name to `n` as it takes into account a possible `n < nmax` situation.

hangtime79:
Looping back in as this was discussed in #112. @tklebel completed this three months ago. Hoping @hadley will get a chance to act on this pull request soon. Until then those who need it can use @tklebel version.

torbjorn:
Is there any reason why this feature change needs to take a year to complete?

torbjorn:
And not sure if this has been covered, but why not allow all lines to be read before guessing the type? perhaps with n=-1 or n=0?

jennybc:
Will live on in #257.

Issue #169, opened 2016-05-01
Unexpected error specifying col names/types in a file with a blank column


MichaelChirico:
I've got an excel file with, say 10 columns.

The first column is blank, and `read_excel` intelligently does away with this column, so that

```
data <- read_excel("file.xls")
ncol(data)
```

is 9, not 10.

It seems natural, then, that I should supply 9 column names to `col_names`, but this is an error, since the processing of column names & types is done before the column disappears. 

I guess this is closely related to #81.

jennybc:
Yes agree this is covered by #81.

Issue #81, opened 2015-04-17
Error using col_names and col_types


meyera:
Hi Hadley,
thanks for another awesome package.
I would like to let you know the following behavior, which seems to be a bug:

``` r
> read_excel(data_file, sheet = "ES") %>% colnames
 [1] "Lfd. Nr."                             "Name"                                 "Vorname"                             
 [4] "Geb.-datum"                           "OP-Datum"                             "Alter"                               
 [7] "Konv. OP-Risiko"                      "Card. Schock"                         "Rescue"                              
[10] "aEuroScore"                           "loEuroScore"                          "EuroSCORE II"                        
[13] "STS_Mortalitaetsrisiko"               "STS_Morb./Mort."                      "STS Long Stay"                       
[16] "STS Stroke"                           "STS Pronl. Vent."                     "STS Ren Fail"                        
[19] "Syntax Score"                         "Rhythmus"                             "A-fib"                               
[22] "PM"                                   "Z.n. AKE"                             "Z.n. ACVB"                           
[25] "Z.n. MKE/R"                           "Z.n. anderer OP (Perikarderoeffnung)" "Stroke, neurol. Defizit"             
[28] "pAVK"                                 "COPD"                                 "SPAP>50"                             
[31] "pHTs"                                 "pHTd"                                 "pHTm"                                
[34] "Diab mell"                            "NI"                                   "KHK"                                 
[37] "Z.n.Stent"                            "Dialyse"                              "Risikofaktoren"  
```

Now I would like to use my own column type and name specification for the 39 excel columns:

``` r
# shorthands for readxl package
num <- "numeric"; txt = "text"; date = "date"; blank = "blank" 
colspec.es <- c(
  id = num,
  es_name = txt,
  es_vorname = txt,
  es_dob = date,
  es_date_op = date,
  es_age = num,
  es_konv_risk = txt,
  es_card_shock = num,
  es_rescue_indication = num,
  es_addESI= num,
  es_logESI= num,
  es_ESII= num,
  es_STS_mort= num,
  es_STS_mort_morb= num,
  es_STS_longstay= num,
  es_STS_stroke= num,
  es_STS_prolong_vent= num,
  es_STS_ren_fail= num,
  es_syntax = num,
  es_rhythm = txt,
  es_afib = num,
  es_pm = num,
  es_sp_ake = num,
  es_sp_cabg = num,
  es_sp_mk_op = num,
  es_sp_other_pericard_open = num,
[...]

PR #170, opened 2016-05-06
Partially addresses #169 by warning for blank columns

MichaelChirico:
Barring a good solution to #81 and #169, this will at least tell the user which columns are being considered as blank so they can adjust their `col_names`/`col_types` accordingly.


hadley:
Thanks - but we'll fix as part of a large colspec cleanup.

PR #175, opened 2016-05-17
Import tibble

krlmlr:
- explicitly test return values
- had to adapt existing tests because of tibble semantics


hadley:
Thanks!

PR #188, opened 2016-07-04
Fix DEFINEDNAME print out when reading Excel files with a name range

PedramNavid:
Wrapped the printing of DEFINEDNAME to prevent printing DEFINEDNAME to the console when reading Excel files with defined name ranges.


jennybc:
Thanks!

Issue #181, opened 2016-06-14
missing va_end/va_start in xlstool.c


BillDunlap:
The non-MSDN version of aspriintf in readxl/src/xlstool.c uses the stdarg va_start and va_end macros at about line 172:

```
va_list ap;

va_start(ap, format); 

i = vsnprintf(NULL, 0, format, ap) + 1;
str = (char *)malloc(i);
i = vsnprintf(str, i, format, ap);

va_end(ap);
```

There should be a va_end(ap) after the first call to vsnprintf() and a va_start(ap, format) before the second.  Otherwise, depending on how the main program is linked, the second call to vsnprintf
may get garbage arguments from the variadic argument list to asprintf.

PR #195, opened 2016-08-05
Fixed invalid use of va_list

rohan-shah:
It's not standards compliant to use a va_list in more than one call. Unfortunately this works on _some_ platforms.

See http://stackoverflow.com/questions/9309246/repeated-use-of-a-variadic-function-argument-doesnt-work

### DIFF
--- a/src/xlstool.c
+++ b/src/xlstool.c
@@ -169,13 +169,14 @@ static int asprintf(char **ret, const char *format, ...)
 	int i;
     char *str;
 
-	va_list ap;
+	va_list ap, ap2;
 
 	va_start(ap, format); 
+	va_copy(ap2, ap);
 
 	i = vsnprintf(NULL, 0, format, ap) + 1;
 	str = (char *)malloc(i);
-	i = vsnprintf(str, i, format, ap);
+	i = vsnprintf(str, i, format, ap2);
 
 	va_end(ap);
 
--- a/src/xlstool.c
+++ b/src/xlstool.c
@@ -167,17 +167,18 @@ static int asprintf(char **ret, const char *format, ...)
 static int asprintf(char **ret, const char *format, ...)
 {
 	int i;
-    char *str;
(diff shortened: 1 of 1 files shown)


hadley:
Could you please give a bit more context for this change? Does it fix broken code or a compiler warning/error?

rohan-shah:
This code is broken on some platforms. I believe I might have pasted the wrong link in the report. 

From http://man7.org/linux/man-pages/man0/stdarg.h.0p.html:
"The object ap may be passed as an argument to another function; if that function invokes the va_arg() macro with parameter ap, the value of ap in the calling function is unspecified and shall be passed to the va_end() macro prior to any further reference to ap."

kevinushey:
That page also seems to say:

>          Each invocation of the va_start() and va_copy() macros shall be
>        matched by a corresponding invocation of the va_end() macro in the
>        same function.

So, IIUC, we need another `va_end()` call on `ap2`. This StackOverflow post also seems pertinent: http://stackoverflow.com/questions/26953289/is-it-safe-to-pass-a-va-list-to-another-function-without-using-va-copy

rohan-shah:
Oops. I think that should be better? Apparently re-initialization using va_start is allowed.

hadley:
Given that we can't see this problem locally, I really appreciate some thoughts on how to test this.

Alternatively given #215 maybe we can just assume `asprintf()` is always available? Or maybe take a modern implementation wholesale from elsewhere?

jennybc:
Another consideration is that the proposed change targets a file included from libxls.

@hadley What's the policy on that?

jennybc:
This is also #181.

hadley:
libxls is basically unmaintained so I don't expect to pull in any upstream changes. Patching the source directly is fine in this case (although we do want to avoid in general)

jennybc:
Thanks!

PR #210, opened 2016-10-11
More helpful error for blank issue in #81 and #169

ecortens:
This pull request adjusts the error message returned when the length of col_types does not match what read_excel() expects, which often occurs when there are blank columns. While it doesn't solve #81 or #169, it does provide the user helpful diagnostics, enabling them at least to know how many columns read_excel() thinks the file has.

### DIFF
--- a/src/XlsWorkSheet.h
+++ b/src/XlsWorkSheet.h
@@ -90,7 +90,7 @@ public:
   Rcpp::List readCols(Rcpp::CharacterVector names, std::vector<CellType> types,
                       std::string na, int nskip = 0) {
     if ((int) names.size() != ncol_ || (int) types.size() != ncol_)
-      Rcpp::stop("Need one name and type for each column");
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Hey, thanks! Have you seen my #170?

ecortens:
Indeed, I did see #170. That works too! Anything to avoid trying to guess how many blank columns read_excel() has found.

MichaelChirico:
Perhaps the combination of these two is warranted -- this covers cases when it may be just a user error passing the wrong number of columns; mine is specifically for pointing out that blanks were found.

ecortens:
Agreed. Good call!

MichaelChirico:
Just need @hadley to weigh in =D

hadley:
We'll fix this as part of the general colspec cleanup - thanks!

PR #215, opened 2016-11-07
Assume all linux flavors have asprintf

### LINKED ISSUE #189: Problems with reading xls file on Centos
Hello,

I have noticed a problem when using `read_excel` for xls files on Centos OS (using readxl 0.1.1).
The column names, columns of character objects and blank columns are not read in.

An example of this is when running your test test-missing-values.R, and all the tests for the xls files fail on my OS. Using the `read_excel` function for the file missing-values.xls, it returns;

``` r
     structure(c("NA", "8", "8"), class = "AsIs")
1 NA                                           NA
2  1                                            8
3  1                                            8
```

Also for the file empty-named-column.xls, it returns;

``` r
     structure(c(" 2", "NA", "NA"), class = "AsIs")
1 1                                              2
2 2                                             NA
3 3                                             NA
```

Similar behaviour is shown for the files utf8-sheets.xls and blanks.xls.

However, for the files dates-1900.xls and dates-1904.xls, the function works correctly, but I believe this is because it has no column names, blank columns or character objects.

This issue is similar to that mentioned in #128 and #180.

``` r eval=FALSE
> sessionInfo()
R version 3.2.5 (2016-04-14)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: CentOS release 6.3 (Final)

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8       
 [4] LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] tools     stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] validation.readxl_0.1.1   validationTools_0.2.0-166 png_0.1-7                
 [4] knitr_1.12.3              XML_3.98-1.4              RUnit_0.4.31             
 [7] roxygen2_5.0.1            devtools_1.10.0           readxl_0.1.1             
[10] testthat_0.11.0          

[1] Rcpp_0.12.4   digest_0.6.9  crayon_1.3.1  withr_1.0.1   magrittr_1.5  stringi_1.0-1 stringr_1.0.0
[8] memoise_1.0.0
```

gergness:
Fixes #189 and doesn't seem to have negative impacts on other platforms (I used r-hub to test Debian, Fedora, Ubuntu, CentOS and Windows).

It seems like we might be able to delete the if statement altogether, but this change seemed less likely to cause problems.

### DIFF
--- a/src/unix/config.h
+++ b/src/unix/config.h
@@ -2,7 +2,7 @@
 /* config.h.in.  Generated from configure.in by autoheader.  */
 
 /* Define to 1 if you have the `asprintf' function. */
-#if defined(_GNU_SOURCE) || defined(_DEFAULT_SOURCE) || defined(__APPLE__) || defined(__sun)
+#if defined(_GNU_SOURCE) || defined(_DEFAULT_SOURCE) || defined(__APPLE__) || defined(__sun) || defined(__gnu_linux__)
 #define HAVE_ASPRINTF 1
 #endif
 


jeroenooms:
@hadley Do you remember why you changed the `ASPRINTF` define in  https://github.com/tidyverse/readxl/commit/74596a6cb92b1821bee1f0f5b7b1ef06a3842dee, https://github.com/tidyverse/readxl/commit/6f8304c179b656fdc8330729cf52a0a51521ee63 and https://github.com/tidyverse/readxl/commit/5a5ababc3a11eb434ab105c0cde17bfbdcd2e585? It seems that asprintf is available on every unix system we target (linux, bsd, solaris, mac).

hadley:
@jeroenooms I have no recollection. I'm happy to do whatever you think is best.

PR #235, opened 2017-01-26
Refactor sheet initialization

jennybc:
This will not get merged but I want to see if this first step looks as you expected. If/when it does, I can move on to modifying `colTypes()`, `colNames()`, and `readCols()` accordingly.

### DIFF
--- a/src/XlsxCell.h
+++ b/src/XlsxCell.h
@@ -36,12 +36,13 @@ class XlsxCell {
 
 public:
 
-  XlsxCell(rapidxml::xml_node<>* cell): cell_(cell) {
+  XlsxCell(rapidxml::xml_node<>* cell, int row = -1, int col = -1): cell_(cell) {
     rapidxml::xml_attribute<>* ref = cell_->first_attribute("r");
-    if (ref == NULL)
-      Rcpp::stop("Invalid cell: lacks ref attribute");
-
-    location_ = parseRef(ref->value());
+    if (ref == NULL) {
+      location_ = std::make_pair(row, col);
+    } else {
+      location_ = parseRef(ref->value());
+    }
   }
 
   int row() {
--- a/src/XlsxWorkSheet.h
+++ b/src/XlsxWorkSheet.h
@@ -20,6 +20,7 @@ class XlsxWorkSheet {
   rapidxml::xml_document<> sheetXml_;
   rapidxml::xml_node<>* rootNode_;
   rapidxml::xml_node<>* sheetData_;
+  std::vector<XlsxCell> cells_;
   int ncol_, nrow_;
 
 public:
@@ -41,6 +42,7 @@ public:
     if (sheetData_ == NULL)
       Rcpp::stop("Invalid sheet xml (no <sheetData>)");
 
+    loadCells();
     cacheDimension();
   }
 
@@ -65,7 +67,6 @@ public:
       }
     }
   }
-
 
   std::vector<CellType> colTypes(const StringSet& na, int nskip = 0, int n_max = 100, bool has_col_names = false) {
     rapidxml::xml_node<>* row = getRow(nskip + has_col_names);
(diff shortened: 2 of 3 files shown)


src/XlsxWorkSheet.h
```diff
@@ -216,47 +217,69 @@ class XlsxWorkSheet {
...
   void computeDimensions() {
-    // If <dimension> not present, iterate over all rows and cells to count
     nrow_ = 0;
     ncol_ = 0;
+    typedef std::vector<XlsxCell>::size_type vec_sz;
+    vec_sz n_cells = cells_.size();
 
-    for (rapidxml::xml_node<>* row = sheetData_->first_node("row");
-      row; row = row->next_sibling("row")) {
+    for (vec_sz i = 0; i != n_cells; ++i) {
```
hadley: Can you replace this for loop with an iterator? (I can provide more details if needed)

src/XlsxCell.h
```diff
@@ -36,12 +36,13 @@ class XlsxCell {
 
 public:
 
-  XlsxCell(rapidxml::xml_node<>* cell): cell_(cell) {
+  XlsxCell(rapidxml::xml_node<>* cell, int row = -1, int col = -1): cell_(cell) {
```
hadley: The negative values are just there so the code compiles, right? In the long run these will be required.

src/XlsxWorkSheet.h
```diff
@@ -20,6 +20,7 @@ class XlsxWorkSheet {
   rapidxml::xml_document<> sheetXml_;
   rapidxml::xml_node<>* rootNode_;
   rapidxml::xml_node<>* sheetData_;
+  std::vector<XlsxCell> cells_;
```
hadley: One thing to bear in mind with this change is that now XlsxWorkSheet becomes a fairly heavy object - that means we need to be more careful about how it is passed around from function to function. I don't think there's anything you need to do right now, but at some point we'll need to review the creation and use of `XlsxWorkSheet` more widely.

src/XlsxWorkSheet.h
```diff
@@ -216,47 +217,69 @@ class XlsxWorkSheet {
...
   }
 
+  void loadCells() {
+    rapidxml::xml_node<>* row = sheetData_->first_node("row");
+    if (row == NULL) {
+      Rcpp::stop("Invalid sheet xml: no <row>");
+    }
+
+    int j = 0;
+    for (int i = 0; row; row = row->next_sibling("row")) {
```
hadley: This for loop seems a bit jumbled to me because you're iterating over two things in parallel but the initialisation and updates are in different places. Maybe move `int i = 0` outside of the loop? And `int j = 0` inside the loop?

src/XlsxWorkSheet.h
```diff
@@ -216,47 +217,69 @@ class XlsxWorkSheet {
...
+      j = 0;
+      for (rapidxml::xml_node<>* cell = row->first_node("c");
```
hadley: Do we need to check that `cell` exists here? (i.e. should we use the same pattern as `row` above?

Also should we stick the convention that we give the C++ object the same name as the xml node?

jennybc: The inner loop over cells is taken verbatim from the old `computeDimensions()`, so I didn't make an active decision to skip the check or depart from a naming convention.

Re: checking. I've never seen an xlsx where you get this far and can't find a cell. IME the inability to find a cell node will come from use of nonstandard namespace prefixes. If that's the case, it will trigger an early exit much earlier, such as failure to find the enclosing row node.

Should I check I can find at least one cell outside the loop, with the check for `row`?

src/XlsxWorkSheet.h
jennybc: OK. This was me wishing I could do this, which I apparently I can't, and finding some compromise:

``` c++
for (int i = 0; row; row = row->next_sibling("row")) {
      for (int j = 0, rapidxml::xml_node<>* cell = row->first_node("c");
      ...
  }
}
```

Since `j  = 0` happens once per row, I thought I should define it outside the loop. But if it's OK to put `int j = 0` inside the loop, I will do that.

src/XlsxWorkSheet.h
jennybc: I have seen them, so will have a try.

src/XlsxCell.h
jennybc: If this was R, I would set them to `NA_integer_`. It feels like you should be able to create an XlsxCell w/o having to guess at its location. I wanted valid placeholders for row and col that are not garbage but that are obviously nonsense. Is there a better way?

jennybc:
This effort now picks up in #240.

PR #238, opened 2017-01-26
Do not use variadic macros

### LINKED ISSUE #176: If using CXX1XSTD should also use CXX1X compiler
readxl sets PKG_CXXFLAGS to include CXX1XSTD (added in commit https://github.com/hadley/readxl/commit/c957ce154a31713f400e348ca30034a66bfcd9be), but does not use the CXX1X compiler.

If the user speficies an alternative compiler for CXX1X (and CXX1XSTD) in Makevars this can result in readxl failing to build (see https://travis-ci.org/leeper/rio/builds/131807863#L1641-L1645 for an example).

In addition the comment confuses me, if the issue is compiling C code with the C99 standard as the submission comments suggest, should we not be setting `PKG_CFLAGS=-std=c99` instead?

jimhester:
They are a C99 extension which then requires =c++11 / c++0x causing
other issues (#176) as we do not actually use the CXX1X compiler.

The only use of fprintf in ole.c uses two additional arguments, so we
will just hard code that number directly.

Fixes #176


jennybc:
Thanks!

PR #239, opened 2017-01-26
Test directory hygiene

jennybc:
@hadley I think this is non-controversial, yes?

It makes it much easier to get at the test specimens during development and the collection is growing.


hadley:
Yeah, the only controversial thing is the use of rprojroot instead of `testthat::test_path()`, but I have a vague memory that there's some good reason that you can't do it

tests/testthat.R
```diff
@@ -1,4 +1,10 @@
 library(testthat)
 library(readxl)
 
-test_check("readxl")
+if (requireNamespace("rprojroot")) {
```
hadley: I think you can just skip this check - it's fine to assume that suggested packages are installed for testing

jennybc:
> Yeah, the only controversial thing is the use of rprojroot instead of `testthat::test_path()`, but I have a vague memory that there's some good reason that you can't do it

At one time, `testthat::test_path()` did not work during `R CMD check`. But I just tried it and it works now, presumably due to https://github.com/hadley/testthat/commit/c83aba9169aa0c6ac3494dcb7a28fefa28ec051a. So once that's in CRAN version of testthat, I can switch over. I will leave myself a reminder.

tests/testthat.R
jennybc: Duly noted. For this and in future.

PR #240, opened 2017-01-30
Load cells at xlsx worksheet ingest; handle skipping and/or blank row…

### LINKED ISSUE #163: free() on invalid pointer
Trying to read [this](https://www.dropbox.com/s/2qjgrxxlx06bxgv/exp.xlsx?dl=0) file with readxl produces the following:

```
> readxl::read_excel("exp.xlsx")
*** Error in `/usr/lib/R/bin/exec/R': free(): invalid pointer: 0x00000000028ccaa8 ***
Aborted
```

The [valgrind report](https://gist.github.com/wrathematics/53c761b3e8cdfce18e3c) doesn't look very useful to me, but I haven't taken a close look at the codebase.

The data is apparently just some random exponential, but was generated by and exported from JMP (it's for a homework problem, so blame my professor, not me). The exported file might be mal-formed somehow, but maybe not.  As I mentioned on twitter, the clang static analyzer `scan-build` (which I use to compile everything) throws several significant warnings when building readxl, though they appear to be coming from libxls.

### LINKED ISSUE #222: Error on reading empty worksheet
Dear Mr.
I am trying to read sheets from xls file

```R
sfile<-"EIU_Data.xls"
sheetlist<-readxl::excel_sheets(sfile)
for (i in 1:length(sheetlist)){
  temp<-readxl::read_excel(sfile,sheet=i,col_names=FALSE)
if (nrow(temp)!=0){
# rest of my own code
}}
```

However I get this error, since the last 3 sheets are empty.
Error in xls_cols(path, sheet, col_names = col_names, col_types = col_types,  : 
  `col_names` and `col_types` must have the same length
Is there a way to solve this problem

[EIU_Data.zip](https://github.com/hadley/readxl/files/675641/EIU_Data.zip)

### LINKED ISSUE #224: n blank rows in xlsx lead to n trailing NA rows in data frame
Moving from #203, which is about a different problem with blank cells.

Comment I deleted there from @burchill:

> `readxl` gets the initial dimensions of the worksheet, which it thinks **ALWAYS** starts at `A1`, and then goes to whatever the maximum row and column are.  
>
> But `.xlsx` files don't store empty rows, so if the first few rows are entirely empty, `readCols` starts from the first **NON-EMPTY** row, but then still iterates until it gives values for every cell in a matrix of the dimensions it originally measured,   You can replicate it with any simple `.xlsx` file where the data doesn't start until a few rows down.  E.g.:

A | B | C
--|--|--
 | | 
 | | 
 | | 
val1 | val2 | val3
val4 | val5 | val6

becomes:

val1 | val2 | val3
--|--|--
val4 | val5 | val6
`NA` | `NA` | `NA` 
`NA` | `NA` | `NA` 
`NA` | `NA` | `NA`

jennybc:
…s; fixes #224, fixes #163 

Partially closes #222 (skipping), which will remain open to cover re: xls. Other linked issues are mentioned in NEWS and commit messages.


src/CellType.h
```diff
@@ -142,6 +142,7 @@ inline Rcpp::RObject makeCol(CellType type, int n) {
   switch(type) {
   case CELL_BLANK:
     return R_NilValue;
+    break;
```
hadley: Doh

src/XlsxWorkSheet.cpp
```diff
@@ -47,6 +47,11 @@ List read_xlsx_(std::string path, int sheet, RObject col_names,
 
   XlsxWorkSheet ws(path, sheet);
 
+  if (ws.nrow() == 0 && ws.ncol() == 0) {
+    Rcpp::List empty(0);
+    return empty;
```
hadley: You could just do `return Rcpp::List(0)`  

(And I assume this gets tibble-ised elsewhere?)

src/XlsxWorkSheet.cpp
```diff
@@ -58,7 +63,7 @@ List read_xlsx_(std::string path, int sheet, RObject col_names,
   {
     sheetHasColumnNames = as<bool>(col_names);
     if (sheetHasColumnNames) {
-      colNames = ws.colNames(nskip);
+      colNames = ws.colNames(std::max(ws.min_row(), nskip));
```
hadley: What's the logic here?

src/XlsxWorkSheet.cpp
```diff
@@ -72,11 +77,16 @@ List read_xlsx_(std::string path, int sheet, RObject col_names,
     Rcpp::stop("`col_names` must be a logical or character vector");
   }
 
+  // if we know there's no data reading to be done, we should return a
+  // 0 row tibble HERE with the correct colNames
+  // TO THINK: case where colTypes were explicitly specified but no data to read
```
hadley: I think you could return a data frame with a 0-length vector for each col, but it's not necessary and few people will notice the omission

src/XlsxWorkSheet.cpp
hadley: Since you use it in a few places, maybe assign to an informatively named variable?

src/XlsxWorkSheet.h
```diff
@@ -16,11 +16,14 @@
 
 class XlsxWorkSheet {
   XlsxWorkBook wb_;
+  // JB QUESTION: seems like a worksheet should know its own name (and/or
+  // position). This could be used in messaging, such as, "SHEET has no data"
```
hadley: Yes, I think that's a good idea

src/XlsxWorkSheet.h
```diff
@@ -52,6 +56,13 @@ class XlsxWorkSheet {
     return nrow_;
   }
 
+  int min_row() {
```
hadley: `const`?

src/XlsxWorkSheet.h
```diff
@@ -52,6 +56,13 @@ class XlsxWorkSheet {
...
+    return min_row_;
+  }
+
+  // JB: this seems to be leftover from previous development?
```
hadley: I think you can/should kill it now that we have `loadCells()`

src/XlsxWorkSheet.h
```diff
@@ -66,31 +77,53 @@ class XlsxWorkSheet {
...
-    int i = 0;
-    while(i < n_max && row != NULL) {
-      for (rapidxml::xml_node<>* cell = row->first_node("c");
-           cell; cell = cell->next_sibling("c")) {
-
-        XlsxCell xcell(cell);
-        if (xcell.col() >= ncol_)
-          continue;
+    // go to row nskip or, if impossible, the next row that exists
+    std::vector<XlsxCell>::const_iterator it = getNextRow(nskip);
```
hadley: I think this should get moved out too. Maybe `loadCells()` should record which is the first non-blank cell that it sees?

src/XlsxWorkSheet.h
```diff
@@ -66,31 +77,53 @@ class XlsxWorkSheet {
...
+    int base = it->row();
+    // advance past column names and any embedded blank rows
+    if (has_col_names) {
+      it = getNextRow(base + 1);
```
hadley: It seems a bit inefficient to start again from scratch to find the next row. Maybe `getNextRow()` should take a ` std::vector<XlsxCell>::const_iterator` and a number of rows to advance?

src/XlsxWorkSheet.h
```diff
@@ -99,54 +132,79 @@ class XlsxWorkSheet {
...
-
     Rcpp::CharacterVector out(ncol_);
-    for (rapidxml::xml_node<>* cell = row->first_node("c");
-         cell; cell = cell->next_sibling("c")) {
-      XlsxCell xcell(cell);
-
-      if (xcell.col() >= ncol_)
-        continue;
+    // go to row nskip or, if impossible, the next row that has data
+    std::vector<XlsxCell>::const_iterator it = getNextRow(nskip);
```
hadley: So does `nskip` include or exclude completely blank rows?

src/XlsxWorkSheet.h
hadley: And maybe also record the first cell after all the skipping?

src/XlsxWorkSheet.h
```diff
@@ -201,61 +261,90 @@ class XlsxWorkSheet {
...
-
-    for (rapidxml::xml_node<>* row = sheetData_->first_node("row");
-      row; row = row->next_sibling("row")) {
+  void loadCells() {
+    rapidxml::xml_node<>* row = sheetData_->first_node("row");
+    if (row == NULL) {
+      // JB: it feels like I should reveal sheet ?name? here
+      // currently worksheet doesn't know it's own name, though
+      // Or perhaps I should say nothing at all?
+      Rcpp::warning("No <row> in sheet xml. No data read.");
```
hadley: I think generally loading an empty file should not generate a warning/message.

src/XlsxWorkSheet.cpp
jennybc: Aha! And YES re: tibble-isation.

src/XlsxWorkSheet.cpp
jennybc: I will see if I can remove or generalize `min_row_` now.

I think it might be wise to record info about where data starts (after advancing past `nskip`) and where the next row starts (assuming we'll need to advance past column names). Because yes I currently do this in at least two different places with the same logic.

src/XlsxWorkSheet.h
jennybc: Agreed and acknowledged in above comment.

src/XlsxWorkSheet.h
jennybc: But you don't know how many rows to advance. However, I think this will all go away anyway, once I record these "milestones" exactly once at the start of worksheet ingest.

src/XlsxWorkSheet.cpp
hadley: Maybe call it `first_row_`? I think that's a bit more descriptive.

And `first_data_row_`?

jennybc: At the time of worksheet construction, right after dimensions are cached, I now mark the first cell with declared row >= `nskip` and the first cell with declared row >= to that of the former. Then all subsequent functions (column names, column types, column reading) just consult that.

src/XlsxWorkSheet.cpp
jennybc: Done.

src/XlsxWorkSheet.h
jennybc: Still need to do this.

src/XlsxWorkSheet.h
jennybc: gone now anyway

src/XlsxWorkSheet.h
jennybc: deleted this

src/XlsxWorkSheet.h
jennybc: Deleted, in favour of marking the cells to start reading from at the time of ingest.

src/XlsxWorkSheet.h
jennybc: warning deleted

src/XlsxWorkSheet.h
jennybc: Done.

src/XlsxWorkSheet.h
jennybc: I do this row identification exactly once now and then just work with the iterators.

jennybc:
@hadley 

I think I've addressed all of your comments and have gotten the NEWs/issues sorted out. Once you are satisfied, I think this can be merged. There are still a few "to do's" but they are best addressed as part of the next phase, where I'll work on column names and types for xlsx.

src/XlsxWorkSheet.h
```diff
@@ -20,28 +20,41 @@ class XlsxWorkSheet {
...
+  std::vector<XlsxCell> cells_;
+  std::string sheetName_;
+  int ncol_, nrow_, nskip_;
+  std::vector<XlsxCell>::const_iterator firstRow_, secondRow_;
 
 public:
 
-  XlsxWorkSheet(XlsxWorkBook wb, int sheet_i): wb_(wb) {
+  XlsxWorkSheet(XlsxWorkBook wb, int sheet_i, int nskip):
+  wb_(wb)
```
hadley: I think our policy is to indent default constructors. @jimhester what do you do?

src/XlsxWorkSheet.h
```diff
@@ -52,101 +65,153 @@ class XlsxWorkSheet {
     return nrow_;
   }
 
-  void printCells() {
-    for (rapidxml::xml_node<>* row = sheetData_->first_node("row");
-         row; row = row->next_sibling("row")) {
+  std::string sheetName() {
```
hadley: `const` ?

tests/testthat/test-empty.R
```diff
@@ -0,0 +1,29 @@
...
+test_that("sheets with column names only are handled", {
+  out <- read_excel(test_sheet("empty-sheets.xlsx"), "header_only")
+  expect_identical(out, tibble::tibble(var1 = numeric(), var2 = numeric()))
+})
+
+test_that("non-empty sheets act that way if we skip past everything", {
+  ## currently we get a 0-row tibble with 2 unnamed columns
+  ## a big improvement but when I rationlize column names, see why these
+  ## empty column names aren't causing the columns to be dropped and fix it
+  out <- read_excel(test_sheet("skipping.xlsx"), skip = 10)
```
hadley: To make it less likely you'll forget about this column, you can use `skip()` to get a message in the tests

src/XlsxWorkSheet.h
jimhester: I [indent them](https://github.com/rstats-db/odbc/blob/71da07fdf51b30043edaf94c094f2da497fae882/src/odbc_result.h#L31-L36).

It might be worth standardizing on [clang-format](https://clang.llvm.org/docs/ClangFormat.html) and setting up a [`.clang-format`](https://github.com/lexicalunit/nanodbc/blob/master/.clang-format) file, which would give us standardized formatting across projects.

src/CellType.h
jimhester: These breaks aren't actually needed, if the function is going to return first you don't need to break out of the switch.

hadley: Either way it seems like we should consistently use or not use in a given switch.

jennybc: I wondered why there had been no suffering due this missing `break`. I will leave it because at least it's now all consistent.

src/XlsxWorkSheet.h
jennybc: This seems not to be what RStudio wants to do. Does it do so for either of you and I've got some setting wrong?

I'm worried about doing it manually, because I use "reindent lines" so often, I fear my indenting will be worse overall, i.e. not internally consistent.

@jimhester Would RStudio implement a ` .clang-format` file?

src/XlsxWorkSheet.h
jennybc: Is there any reason to not do same for `ncol()` and `nrow()`, just above this?

I added for all 3, so speak up if it's a mistake to add for the other two.

tests/testthat/test-empty.R
jennybc: Done! Nice idea.

PR #247, opened 2017-02-06
Use tibble::repair_names() to eliminate empty, NA, duplicated column names

jennybc:


R/read_excel.R
```diff
@@ -57,11 +56,11 @@ read_xls <- function(path, sheet = 1, col_names = TRUE, col_types = NULL,
     col_types <- xls_col_types(path, sheet, na = na, nskip = skip, has_col_names = has_col_names)
   }
 
-  as_data_frame(
+  tibble::repair_names(tibble::as_data_frame(
```
hadley: Maybe switch to `as_tibble()` now too?

jennybc: I should add `tibble (>= 1.0)` anyway, because that's when `repair_names()` appeared.

If I switch to `as_tibble()`, it will need to be `tibble (>= 1.1)`.

Issue #157, opened 2016-01-13
Empty columns: to drop or not to drop?


huftis:
When importing Excel files, empty columns seem to be deleted if one uses a non-zero `skip` argument in `read_excel()`. Here’s an example file:

http://huftis.org/nedlasting/excel/row-col-test.xlsx

The file contains two similar worksheets/tabs. Both have 8 columns (named A–H) and 8 rows. The diagonal is filled with the numbers 1 to 8 for the first worksheet. The second worksheet is similar, but is missing the number 4 and 7, resulting in column 4 and and 7 being empty.

Reading the first worksheet works fine:

```
> read_excel("row-col-test.xlsx", col_names = FALSE, sheet=1, skip=2)
  X1 X2 X3 X4 X5 X6 X7
1  2 NA NA NA NA NA NA
2 NA  3 NA NA NA NA NA
3 NA NA  4 NA NA NA NA
4 NA NA NA  5 NA NA NA
5 NA NA NA NA  6 NA NA
6 NA NA NA NA NA  7 NA
7 NA NA NA NA NA NA  8
```

But reading the second worksheet results in a data frame where the empty columns are missing:

```
> read_excel("row-col-test.xlsx", col_names = FALSE, sheet=2, skip=2)
  X1 X2 X4 X5 X7
1  2 NA NA NA NA
2 NA  3 NA NA NA
3 NA NA  5 NA NA
4 NA NA NA  6 NA
5 NA NA NA NA  8
6 NA NA NA NA NA
7 NA NA NA NA NA
```

Note that the (automatically generated) column names indicate that there once was an X3 and X6 column.

If one uses `skip=0`, empty columns are _not_ dropped:

```
> read_excel("row-col-test.xlsx", col_names = FALSE, sheet=2, skip=0)
    X0   X1   X2   X3   X4   X5   X6   X7
1    A    B    C    D    E    F    G    H
2    1 <NA> <NA> <NA> <NA> <NA> <NA> <NA>
3 <NA>    2 <NA> <NA> <NA> <NA> <NA> <NA>
4 <NA> <NA>    3 <NA> <NA> <NA> <NA> <NA>
5 <NA> <NA> <NA> <NA>    5 <NA> <NA> <NA>
6 <NA> <NA> <NA> <NA> <NA>    6 <NA> <NA>
7 <NA> <NA> <NA> <NA> <NA> <NA> <NA>    8
8 <NA> <NA> <NA> <NA> <NA> <NA> <NA> <NA>
[...]

PR #248, opened 2017-02-06
Don't load empty cells; use empirical worksheet dimensions

jennybc:
There can be cells that look empty to the human eye, but that are styled, e.g. someone applied a custom number format to the cell at some point. I propose we do not load those anymore. There's nothing to learn from their XML.

If such a cell still falls into the target rectangle, it will simply be `NA` in the output (ok, won't be totally true until we stop dropping unnamed, empty columns #157).

If such a cell falls outside the target rectangle, this PR prevents the creation of trailing row(s) or column(s) consisting entirely of `NA`.

In addition to rows/columns of `NA`, this has been a puzzling source of errors about column names and types not being compatible.

To fully fix this, you also can't trust a worksheet's declared dimensions, because these cells "count". Therefore I propose we always compute dimension ourselves.

I haven't dealt with xls yet. I can leave this open and work on that. Or we could resolve this for xlsx and I'd open an issue to fix it for xls.

This branch builds off #247. I'm assuming it will be merged first.


src/XlsxWorkSheet.h
```diff
@@ -296,16 +267,20 @@ class XlsxWorkSheet {
       int j = 0;
       for (rapidxml::xml_node<>* cell = row->first_node("c");
            cell; cell = cell->next_sibling("c")) {
-        XlsxCell xcell(cell, i, j);
-        cells_.push_back(xcell);
+        // don't load a cell with no child nodes, e.g. it only has style
+        rapidxml::xml_node<>* first_child = cell->first_node(0);
```
hadley: I think the check here probably eliminates a check somewhere else.

tests/testthat/test-empty.R
```diff
@@ -28,3 +28,31 @@ test_that("non-empty sheets act that way if we skip past everything", {
...
+test_that("empty cells with a style are not loaded", {
+  ## what's important about this sheet?
+  ## contains empty cells with a custom format
+  ## therefore they appear in the xml and have a style attribute
+  ## where are they?
+  ## in the embedded empty columns, w/ and w/o a name
+  ## in a trailing empty column
+  ## in some trailing rows
+  out <- read_excel(test_sheet("style-only-cells.xlsx"))
+  expect_equal(
```
hadley: I think code like this is slightly cleaner if you define the tibble outside the test

tests/testthat/test-empty.R
```diff
@@ -28,3 +28,31 @@ test_that("non-empty sheets act that way if we skip past everything", {
...
+                   var3 = c("aa", "bb", "cc"),
+                   var5 = c(1, 2, 3))
+  )
+  skip("revisit this when dust settles re: treatment of empty columns")
+})
+
+test_that("user-supplied column names play nicely with empty columns", {
+  skip("waiting for dust to settle re: treatment of empty columns")
+  ## do stuff like this:
+  out <- read_excel(test_sheet("style-only-cells.xlsx"),
```
hadley: House style is

```R
read_excel(
  test_sheet("style-only-cells.xlsx"),
  col_names = LETTERS[1:4]
)
```

src/XlsxWorkSheet.h
jennybc: There's not an obvious one, but I'll keep my eyes peeled.

Styled empty cells were a major source of "cell lacks the t attribute" here, which is why they always lead to numeric `NA` columns. But it's not clear they are the only source? When I get to column types, I'll try to settle this definitively.

https://github.com/tidyverse/readxl/blob/a78440d4f4f4fa62f8b89fe5296a262e9cb24180/src/XlsxCell.h#L122-L124

jennybc: Similar situation with this check for a cell having a `v` node. But I'm not sure if these cells are the only source of this either?

https://github.com/tidyverse/readxl/blob/a78440d4f4f4fa62f8b89fe5296a262e9cb24180/src/XlsxCell.h#L102-L104

tests/testthat/test-empty.R
jennybc: done

tests/testthat/test-empty.R
jennybc: done

PR #253, opened 2017-02-09
Catch "skipping past all the data" ASAP and return 0x0 tibble

jennybc:
If we have skipped past all the data, now this condition will hold and we behave exactly as if the sheet is empty. In particular, we never even manufacture column names or create types.

https://github.com/tidyverse/readxl/blob/fcda8b53c137465cab8e26d72edcf14f66662ff7/src/XlsxWorkSheet.cpp#L45-L47


src/XlsxWorkSheet.h
```diff
@@ -52,9 +52,13 @@ class XlsxWorkSheet {
...
 
     nskip_ = nskip;
+    nrow_ = 0;
+    ncol_ = 0;
     loadCells();
-    computeDimensions();
     markRows();
+    if (firstRow_ != cells_.end()) {
+      computeDimensions();
+    }
```
hadley: This feels a bit magical to me - i.e. it obviously works, but I have no idea why.

Maybe it's now time for a little refactoring of these functions. What if you combined `markRows()` and `computeDimensions()` and give that function an explicit `nskip` parameter? Then that function would be responsible for setting `firstRow_`, `secondRow_`, `nrow_` and `ncol_`, and I think it would a (tiny) performance boost because you'd eliminate one loop through the cells. Once you have `firstRow_` I don't think you need nskip any more, so that could be removed from the class.

PR #256, opened 2017-02-10
Add `col_type = "list"` option

gergness:
This PR adds the ability to read in a column as a list of length-1 vectors, preserving the type given to the cell by excel.

I can imagine a lot of ways to solve this class of problems, but I found it useful when working with an excel worksheet where a bunch of cells were supposed to be of the form "NUMBER-NUMBER" but were converted to dates by excel. 

I also think there should be a warning when readxl encounters excel dates in text or numeric columns. I find it confusing that it silently gives a number, even if the original date is technically recoverable from the number. I'm happy to add that here, or in a separate PR if you prefer.

```r
library(readxl)
library(purrr)
library(lubridate)
library(dplyr)

example <- read_excel("example.xls")

example
#> # A tibble: 3 × 1
#>           var1
#>          <chr>
#> 1        30-30
#> 2        20-20
#> 3 43018.000000

example <- read_excel("example.xls", col_type = "list") %>%
  mutate(var1 = map_if(var1, is.POSIXct, ~paste0(month(.), "-", day(.))) %>% flatten_chr)

example
#> # A tibble: 3 × 1
#>    var1
#>   <chr>
#> 1 30-30
#> 2 20-20
#> 3 10-10
```


hadley:
I have mixed feelings about this because at some point you're better off switching to an approach like tidyxl/linen where you get a tibble with one row per cell.

gergness:
Yeah, I agree that it is pushing the boundaries. I think I was drawing the line at tabular data. 

Closing this because of the conflicts and I'll open an issue about these columns with date types hidden between other types because I still think that readxl could improve.

Re: comment on #261, I think this could be implemented without separating COL_TYPE from CELL_TYPE, but it felt cleaner to split it out since the parser would never return a CELL_LIST. 

Thank you for considering!

jennybc:
I agree, but lean towards merging this. I think it just makes the cut re: how far readxl should go to help with tricky data, before recommended packages more squarely aimed at that.

Separating the notion of a cell type and col type just seems like a good idea anyway.

Then it's just the added overhead of the list col type, which seems tolerable.

> I also think there should be a warning when readxl encounters excel dates in text or numeric columns. I find it confusing that it silently gives a number, even if the original date is technically recoverable from the number.

This is a separate but also interesting question I though about yesterday re: the order of the cell types in the `CellType` enum. @gergness Will you open that thought as an issue?

gergness:
Oh great, will do.

jennybc:
@gergness I just merged something big. Can you either rebase this PR or close and make a new one?

hadley:
If you update, note that skip is a col type, not a cell type

PR #257, opened 2017-02-10
add guess_max param for guessing col_type; fixes #223

### LINKED ISSUE #223: Adopt guess_max a la readr
Propose `read_excel()` follow readr and gain a `guess_max` parameter with similar default behaviour:

> guess_max   Maximum number of records to use for guessing column types.
> ... n_max = Inf, guess_max = min(1000, n_max), ...

Do you approve in principle @hadley?

Would subsume issues #206 #131 , PRs #138 #143

jennybc:
Incorporates and extends @tklebel's PR re: adding a parameter to control number of rows used for col type guessing. I did the rebase locally and closed the original PR.

Goal was to make everything as close to readr as possible. `read_excel()` does not (yet?) have an `n_max` parameter, so that's [a small difference](https://github.com/tidyverse/readr/blob/8e7febc62e0ae4b605e82a528bdd98ac966d9ed5/R/read_delim.R#L96).

Why now? Turns out the column type stuff is easier to work on and test if `guess_max` is exposed, so I got this out of the way.


hadley:
(Most of this diff seems to be a spurious re-ordering of the stuff that Rcpp generates; I've seen this before but I can't remember what causes it)

jennybc:
`parseXml` and `countRows` re-situate themselves perpetually 🙂🙃🙂🙃 and I don't know why.

I asked @jimhester about it but he's not familiar with the problem. I try to NOT commit these but apparently I did this time.

Issue #198, opened 2016-08-25
Feature request: add column guessing as in readr


dapper2dan:
It would be nice to have the same column guessing with the same syntax as in readr.
https://github.com/hadley/readr/releases/tag/v1.0.0

jennybc:
A specific request that appears in #81, at the very least, is to make it easy to set one column type for all columns. Done in readr via, e.g., `cols(.default = "c")`.

jennybc:
It would be nice to specify col type for certain columns and allow others to be guessed.

jennybc:
It would be nice to know whether a col type was guessed or specified by user. Conceivably we would warn less (and maybe even coerce more aggressively?) if the type was requested vs. guessed.

jennybc:
I've edited the title to reflect my current thinking: it would be good to align with readr re: explicit column specification (including skipping and guessing, which is already supported). But I don't ever see doing readr-style column guessing. readr guesses based on data, which it must. readxl guesses based on Excel types, which -- I argue -- it must.

nickbond:
Setting aside the guessing of formats, it would be good to at least align the syntax with readr. At the moment read_excel uses col_types "text" for example, whereas readr functions refer to "character".

llrs:
I asked in #571 to be able to supply a long list of named col_type and use it when the name of the column matches a name on the col_type argument. If I understood correctly readr col_type argument if a column name is not present it doesn't try to guess the type of the column. 

```r
library("readr")
mtcars_csv <- system.file("extdata", "mtcars.csv", package="readr")
col_types_list <- cols_only(
    mpg = col_double(),
    cyl = col_integer(),
    disp = col_double()
)
df1 <- read_csv(mtcars_csv, col_types=col_types_list)
ncol(df1) == 3
df2 <- read_csv(mtcars_csv)
ncol(df2) == 11
```

I think that readxl should try to guess the column types if the type is not provided. Thanks!

TreyRoady:
The ability to specify column types by name is absolutely key for large data sets with many columns.

In my case, I'm importing internal enterprise data reports. Read_excel is currently mangling the date-time objects in it, and the ability to go back and tweak just that column would be absolutely invaluable. Otherwise, I'm having to go back and hand-define hundreds of columns just to redefine one.

jennybc:
This is on the roadmap for my next several months. The plan is to do a major upgrade of the col spec here at the same time as adding similar functionality to googlesheets4.

[...]

PR #261, opened 2017-02-12
Refactor xlsx col type handling

jennybc:
1. If user wants to skip a column, now we say `"skip"` instead of `"blank"`. Created a new `CellType`, `CELL_SKIP` for this. Existing `CellType`, `CELL_BLANK` now really means what it says.

    We're using `CellType` for 2 different concepts that almost coincide but not quite: cell type and column type. I thought about separating them but it didn't seem worth it. A very recent PR (#256) from a user proposes to implement a `"list"` type, where each cell can have the type declared in the Excel. And he, possibly out of necessity, had to separate the cell type from column type. Something to think about.

1. Leading and embedded empty columns are never dropped.

1. If `col_types` has length one, recycle it to have length = number of columns.

1. More effort to make sense of user-specified `col_names`, if there are also user-specified `col_types`. Specifically, if column(s) will be skipped, accept full-length `col_names` or `col_names` post-skip. *I suspect you might not approve of this?*

AWKWARD: a few things had to be fixed or improved. But it doesn't seem to make sense to build up a flexible col spec system in readxl, since it should just work like readr (#198). But that would require pulling lots of logic out of readr into a col spec package. Which isn't imminent. 🤔


src/CellType.h
```diff
@@ -5,10 +5,13 @@
 #include <libxls/xls.h>
 #include "StringSet.h"
 
+// CELL_SKIP  can arise only as a user-specified column type
+// CELL_BLANK can arise only from an individual cell during type guessing
 enum CellType {
   CELL_BLANK,
   CELL_DATE,
   CELL_NUMERIC,
+  CELL_SKIP,
```
hadley: I think it's coincidental that the names were previously in alphabetical order. I think `CELL_SKIP` should go next to `CELL_BLANK`.

src/CellType.h
```diff
@@ -26,11 +29,14 @@ inline std::vector<CellType> cellTypes(Rcpp::CharacterVector x) {
       types.push_back(CELL_DATE);
     } else if (type == "numeric") {
       types.push_back(CELL_NUMERIC);
+    } else if (type == "skip") {
+      types.push_back(CELL_SKIP);
     } else if (type == "text") {
       types.push_back(CELL_TEXT);
     } else {
-      Rcpp::warning("Unknown type '%s' at position %i. Using text instead.",
+      Rcpp::warning("Unknown type '%s' at position %i. Using 'text' instead.",
```
hadley: It might be better to just stop here to clarify that error handling happens on the R side.

src/XlsWorkSheet.h
```diff
@@ -115,12 +119,15 @@ class XlsWorkSheet {
...
           break;
+        case CELL_SKIP:
+          break;
         case CELL_NUMERIC:
           switch(type) {
           case CELL_BLANK:
             REAL(col)[i] = NA_REAL;
             break;
           case CELL_NUMERIC:
+          case CELL_SKIP:
```
hadley: Shouldn't this go with `CELL_BLANK`? (Not that it really matters)

src/XlsxWorkSheet.cpp
```diff
@@ -64,18 +66,63 @@ List read_xlsx_(std::string path, int sheet, RObject col_names,
...
-  // Standardise column types --------------------------------------------------
+  // Get column types --------------------------------------------------
   std::vector<CellType> colTypes;
   switch(TYPEOF(col_types)) {
   case NILSXP:
     colTypes = ws.colTypes(na, guess_max, sheetHasColumnNames);
     break;
   case STRSXP:
     colTypes = cellTypes(as<CharacterVector>(col_types));
+    if ((int) colTypes.size() == 1) {
```
hadley: I think this would be slightly cleaner if you either use `std::fill()` or `Rcpp::rep()`

src/XlsxWorkSheet.cpp
```diff
@@ -64,18 +66,63 @@ List read_xlsx_(std::string path, int sheet, RObject col_names,
...
+    if (colTypes[i] == CELL_BLANK) {
+      colTypes[i] = CELL_NUMERIC;
+    }
+    if (colTypes[i] != CELL_SKIP) {
+      ncol_noskip++;
+    }
+  }
+
+  // Rationalize column names w.r.t. types -----------------------------
+  size_t ncol_names = colNames.size();
```
hadley: Maybe pull this out into a separate helper function? Then you'll be able to re-use for xls.

src/XlsxWorkSheet.h
```diff
@@ -68,64 +68,47 @@ class XlsxWorkSheet {
...
-      while (it != row_end) {
-        XlsxCell xcell = *it;
-        if (xcell.col() < ncol_) {
-          CellType type = xcell.type(na, wb_.stringTable(), wb_.dateStyles());
-          if (type > types[xcell.col()]) {
-            types[xcell.col()] = type;
-          }
+      XlsxCell xcell = *it;
+      if (xcell.col() < ncol_) {
+        CellType type = xcell.type(na, wb_.stringTable(), wb_.dateStyles());
```
hadley: I wonder if it might be worthwhile to compute and cache the type in the cell constructor?

src/XlsxWorkSheet.h
```diff
@@ -68,64 +68,47 @@ class XlsxWorkSheet {
...
-      }
-
-      while (it != row_end) {
-        XlsxCell xcell = *it;
-        if (xcell.col() < ncol_) {
-          CellType type = xcell.type(na, wb_.stringTable(), wb_.dateStyles());
-          if (type > types[xcell.col()]) {
-            types[xcell.col()] = type;
-          }
+      XlsxCell xcell = *it;
```
hadley: This will copy the xcell - I think it would be better to rename `it` to `xcell` and then use (e.g.) `xcell->type()`

src/XlsxWorkSheet.h
```diff
@@ -68,64 +68,47 @@ class XlsxWorkSheet {
...
     }
 
+    // base is row the data starts on **in the spreadsheet**
     int base = firstRow_->row() + has_col_names;
     // we have consulted i rows re: determining col types
     int i;
     // account for any empty rows between column headers and data start
     i = it->row() - base;
+    // m is the max row number seen so far
+    int m = it->row();
```
hadley: Calling these similar things `i` and `m` feels a bit confusing.

tests/testthat/test-missing-values.R
```diff
@@ -70,14 +70,35 @@ test_that("text values in numeric column gives warning & NA", {
...
+test_that("empty (styled) cells are not loaded, but can survive as NA", {
+  ## what's important about this sheet?
+  ## contains empty cells with a custom format
+  ## therefore they appear in the xml and have a style attribute
+  ## where are they?
+  ## in the embedded empty columns, w/ and w/o a name
+  ## in a trailing empty column WHICH SHOULD BE DROPPED
+  ## in some trailing rows WHICH SHOULD BE DROPPED
+  out <- read_excel(test_sheet("style-only-cells.xlsx"))
+  df <- tibble::tibble(var1 = c("val1,1", "val2,1", "val3,1"),
```
hadley: Indenting

hadley:
Wrt 4., I think it's ok, but we should explore whether we can do the same thing in readr. I have a vague memory that it's more complicated because you can associate names with types there.

src/XlsxWorkSheet.h
jennybc: One reason to NOT do this: when cells are constructed in `loadCells()`, we don't know about skipped columns yet.

Currently, we don't read type (or content) of cells in skipped columns (and I've said as much in the docs). We do the bare minimum: register the cell's existence and its coordinates. Everything else is deferred to `readCols()`, which operates with the knowledge in `col_types`.

Does this change your opinion?

hadley: Oh that's a good point and it does change my opinion

src/CellType.h
jennybc: I just got an education on the importance of the order here. Short version: `CELL_BLANK` must be first (= default type), but `CELL_SKIP` can come next.

src/CellType.h
jennybc: done

src/XlsxWorkSheet.cpp
jennybc: Used `std::fill()`. Nice to know about that!

src/XlsxWorkSheet.cpp
jennybc: New helper function = `reconcileNames()`.

src/XlsxWorkSheet.h
jennybc: Simplified. `m` is gone.

src/XlsxWorkSheet.h
jennybc: Done, here and other places.

tests/testthat/test-missing-values.R
jennybc: Done.

jennybc:
This is ready for re-review.

src/XlsxWorkSheet.h
```diff
@@ -149,111 +119,95 @@ class XlsxWorkSheet {
...
-          continue;
-        }
-        CellType type = xcell.type(na, wb_.stringTable(), wb_.dateStyles());
-        Rcpp::RObject col = cols[xcell.col()];
-        // row to write into
-        int row = xcell.row() - base;
-        // Needs to compare to actual cell type to give warnings
-        switch(types[xcell.col()]) {
-        case CELL_BLANK:
+      CellType type = xcell->type(na, wb_.stringTable(), wb_.dateStyles());
```
jennybc: It's a pity this diff looks so big, because it's really not. In this section, it's just the addition(s) of `CELL_SKIP`. Which never comes up, but it seems I have to include all `enum` values in all the `switch()`es.

src/XlsxWorkSheet.cpp
```diff
@@ -89,40 +85,18 @@ List read_xlsx_(std::string path, int sheet, RObject col_names,
...
   for (size_t i = 0; i < colTypes.size(); i++) {
     if (colTypes[i] == CELL_BLANK) {
       colTypes[i] = CELL_NUMERIC;
     }
-    if (colTypes[i] != CELL_SKIP) {
-      ncol_noskip++;
-    }
   }
 
   // Rationalize column names w.r.t. types -----------------------------
```
hadley: I think you can delete this comment now (as the function name should be sufficient)

src/XlsxWorkSheet.h
```diff
@@ -149,30 +136,23 @@ class XlsxWorkSheet {
...
+    while (xcell != cells_.end()) {
 
       if ((i + 1) % 1000 == 0) {
         Rcpp::checkUserInterrupt();
       }
-
-      XlsxCell xcell = *it;
-      if (types[xcell.col()] == CELL_SKIP || xcell.col() >= ncol_) {
-        it++;
+      if (types[xcell->col()] == CELL_SKIP || xcell->col() >= ncol_) {
```
hadley: This would all be a little more compact with `int j = xcell->col()`

src/XlsxWorkSheet.h
```diff
@@ -149,30 +136,23 @@ class XlsxWorkSheet {
...
         continue;
       }
 
-      CellType type = xcell.type(na, wb_.stringTable(), wb_.dateStyles());
-      Rcpp::RObject col = cols[xcell.col()];
+      CellType type = xcell->type(na, wb_.stringTable(), wb_.dateStyles());
+      Rcpp::RObject col = cols[xcell->col()];
       // row to write into
-      int row = xcell.row() - base;
+      int row = xcell->row() - base;
```
hadley: Can you use `i` here?

tests/testthat/test-missing-values.R
```diff
@@ -95,10 +95,11 @@ test_that("empty (styled) cells are not loaded, but can survive as NA", {
...
-                       var2 = NA_real_,
-                       var3 = c("aa", "bb", "cc"),
-                       X__1 = NA_real_,
-                       var5 = c(1, 2, 3))
+  df <- tibble::tibble(
+    var1 = c("val1,1", "val2,1", "val3,1"),
+    var2 = NA_real_,
+    var3 = c("aa", "bb", "cc"),
+    X__1 = NA_real_,
+    var5 = c(1, 2, 3))
```
hadley: Trailing parens should be on its own line

PR #265, opened 2017-02-13
Add ability to read column as a list of length-1 vectors closes #262 …

### LINKED ISSUE #262: Allow user to handle Excel's autoformatting of dates
Excel has aggressive auto-formatting that the user may not have noticed when inputting data. Currently readxl coerces date cells into the raw number used by excel, but usually what I want is the formatted string, as it appears in Excel.

I think that a column with rows that have types that cannot be automatically coerced is a unique to Excel worksheets because of Excel's aggressive auto-formatting. The other tidyverse packages for data import don't have to deal with it, because none of them import data with types that can change across rows (readr doesn't receive type information from the file and haven/feather always receive the same type for a column).

#256 implemented a solution that imported the column as a list type which preserved the type information.

I can also imagine adding a formatting parameter that would give the user control over how a date is formatted when a character column encounters it. I think this would be less flexible, but may make more sense in readxl.

I'm happy to rebase #256 if you are interested, take a stab at implementing the formatting argument, or leave it alone.

gergness:
…(rebase of #256)

Splits out COL_TYPE from CELL_TYPE, as discussed and implemented the list columns.

Thank you!


src/CellType.h
```diff
@@ -5,35 +5,51 @@
...
+      types.push_back(COL_BLANK);
     } else if (type == "date") {
-      types.push_back(CELL_DATE);
+      types.push_back(COL_DATE);
     } else if (type == "numeric") {
-      types.push_back(CELL_NUMERIC);
+      types.push_back(COL_NUMERIC);
     } else if (type == "skip") {
-      types.push_back(CELL_SKIP);
+      types.push_back(COL_SKIP);
```
jennybc: Can you put these in the same order as the ColType enum (blank, date, numeric, text, skip, list) here and everywhere possible? Just to reduce the amount of future wondering whether departures from the standard order are random or are required by the logic.

Can you think of a reason to put "skip" before "list" in the ColType enum? If not, I suspect it should be "list" then "skip"

src/CellType.h
```diff
@@ -42,13 +58,14 @@ inline std::vector<CellType> cellTypes(Rcpp::CharacterVector x) {
...
-inline std::string cellTypeDesc(CellType type) {
+inline std::string colTypeDesc(ColType type) {
   switch(type) {
-  case CELL_BLANK:   return "blank";
-  case CELL_DATE:    return "date";
-  case CELL_NUMERIC: return "numeric";
-  case CELL_SKIP:    return "skip";
-  case CELL_TEXT:    return "text";
+  case COL_BLANK:   return "blank";
+  case COL_SKIP:    return "skip";
```
jennybc: Ditto re: order comment from above.

tests/testthat/test-col-types.R
```diff
@@ -104,4 +104,16 @@ test_that("wrong length col types generates error", {
   )
 })
 
+test_that("list column preserves type [xlsx]", {
```
jennybc: Will you an embedded empty cell in these test sheets and an expectation that it comes in as logical?

BTW, elsewhere readxl defaults to numeric NA for missing data, but the plan is to move to logical, so it's fine that you've already done that for the "list" col type.

R/read_excel.R
```diff
@@ -14,8 +14,10 @@ NULL
 #'   unskipped column.
 #' @param col_types Either `NULL` to guess from the spreadsheet or a character
 #'   vector containing one entry per column from these options: "skip",
-#'   "numeric", "date" or "text". The content of a cell in a skipped column is
-#'   never read and that column will not appear in the data frame output.
+#'   "numeric", "date", "text" or "list". The content of a cell in a skipped column is
+#'   never read and that column will not appear in the data frame output. A list cell
+#'   loads a column as a list of length 1 vectors, preserving the type information from
```
jennybc: Please reword to convey this: the `"list"` col type applies the same guessing logic as `col_types = NULL`, but on a cell-by-cell basis. And then reuse that wording in NEWS.

jennybc:
Do you care to add an example? I just did something very unsatisfying with the existing example sheet `datasets.xlsx` that sort shows off this new feature but not very well. So I think you'd need to add another example sheet, which would be fine. Can you make a small one based on the real sheets (#262) that prompted this PR?

jennybc:
I just added a little example sheet in master with information on Clippy 📎 which would permit this nice example for the `"list"` col type:

``` r
df <- read_excel(readxl_example("clippy.xlsx"), col_types = c("text", "list"))
df
#> # A tibble: 4 × 2
#>                   name      value
#>                  <chr>     <list>
#> 1                 Name  <chr [1]>
#> 2              Species  <chr [1]>
#> 3 Approx date of death <dttm [1]>
#> 4      Weight in grams  <dbl [1]>
df$value
#> [[1]]
#> [1] "Clippy"
#> 
#> [[2]]
#> [1] "paperclip"
#> 
#> [[3]]
#> [1] "2007-01-01 UTC"
#> 
#> [[4]]
#> [1] 0.9
```

src/CellType.h
gergness: Sure, both done. I left places where COL_SKIP and COL_BLANK are treated the same out of order, but otherwise they follow this pattern.

For the second point, I think that the order only matters in 2 places:
1) CellType's order matters for type promotion (https://github.com/tidyverse/readxl/pull/265/files#diff-eead265d07a76e357310dc100290439fL83)

2) as_ColType uses the fact that the types that are in both enums are in the same order (https://github.com/tidyverse/readxl/pull/265/files#diff-5934b87b042ba3f60d59f8e5e214dd85R26)

tests/testthat/test-col-types.R
gergness: Good call, thanks!

gergness:
I don't see the clippy spreadsheet, but would be happy to.

jennybc:
Thanks for all the changes. I'm happy to add the example, so we can set that aside. I'll ask Hadley if he wants to look this over.

jennybc:
> I don't see the clip spreadsheet

Ah, I had pushed to master on my fork but not on the main repo. Done now. We'll sort it out either way.

src/XlsWorkSheet.h
```diff
@@ -167,6 +164,32 @@ class XlsWorkSheet {
             SET_STRING_ELT(col, i, rString);
           }
           break;
+        case COL_LIST:
+          switch(type) {
+          case CELL_BLANK: {
+            Rcpp::as<Rcpp::List>(col)[i] = Rcpp::LogicalVector(1, NA_LOGICAL);
```
hadley: Can you do this in C API style to match the other cases? I think Rcpp just gets in the way here. (I'm happy to provide hints if you're not familiar with R's C API)

gergness: Aha - I couldn't find SET_VECTOR_ELT when I needed it. 

Is this what you meant, or do you think that it could be further streamlined?

src/XlsWorkSheet.h
```diff
@@ -167,6 +164,32 @@ class XlsWorkSheet {
             SET_STRING_ELT(col, i, rString);
           }
           break;
+        case COL_LIST:
+          switch(type) {
+          case CELL_BLANK: {
+            SET_VECTOR_ELT(col, i, Rcpp::LogicalVector(1, NA_LOGICAL));
```
gergness: Cool, thank you. 

BLANK and NUMERIC cell types were wins, but my attempts to change DATE were worse (the class and tz attribute steps are simpler in c++). I'm not sure about TEXT, I did change it some (for xls files), but I'm not sure it's any more clear.

src/XlsWorkSheet.h
```diff
@@ -167,6 +164,30 @@ class XlsWorkSheet {
...
+          case CELL_DATE: {
+            Rcpp::RObject cell_val = Rf_ScalarReal((cell.d - offset_) * 86400);
+            cell_val.attr("class") = Rcpp::CharacterVector::create("POSIXct", "POSIXt");
+            cell_val.attr("tzone") = "UTC";
+            SET_VECTOR_ELT(col, i, cell_val);
+            break;
+          }
+          case CELL_TEXT: {
+            std::string stdString((char*) cell.str);
+            Rcpp::CharacterVector rString = na.contains(stdString) ? NA_STRING : Rf_mkCharLenCE(stdString.c_str(), 1, CE_UTF8);
```
hadley: Shouldn't this be `Rf_mkCharCE(stdString.c_str(), CE_UTF8)`? (Otherwise you'll always get a length 1 string?)

The combination of C++ and C APIs are a bit ugly to my eye, but there's no obvious way to substantially improve them.

gergness: Yep, fixed.

jennybc:
Thanks @gergness.

PR #271, opened 2017-02-24
xls catch up re: changes to sheet geometry and col specification

### LINKED ISSUE #258: Catch up on the xls side
Many fixes and features apply only to xlsx at the moment.

jennybc:
This became something of a monster PR 👹 sorry. What's here, in order of importance:

  1. xls reading was aligned with xlsx.
      - All logic pushed to the C++ side, instead of R. Overall sheet reading function `read_xls_` is almost carbon copy of `read_xlsx_`. Ditto to various degrees for the XlsWorksheet constructor and member functions.
  1. `read_excel_()` worker created so that sheet standardization, `col_names`/`col_types` checks, tibble-ization, and name repair are applied uniformly, but not repeated.
  1. General hygiene. Removed more vestigial code. Remove name inconsistencies that had no significance. Moved stuff between files, renamed files, eliminated files (e.g. xls-specific cell type content moved to new XlsCell.h header, remaining content folded into ColSpec.h, CellType.h deleted).

Closes #258 *still checking if this closes some other xls issues*


src/XlsWorkBook.h
```diff
@@ -21,11 +22,13 @@ class XlsWorkBook {
 
 public:
 
-  XlsWorkBook(std::string path) {
+  XlsWorkBook(const std::string& path)
```
jennybc: It feels like `offset_` and `customDateFormats_` should belong to the **workbook**, as they do on the xlsx side. And should be read right here.

An XlsxWorksheet has its host XlsxWorkbook as a member, which then provides access to this date info. My attempts to do same for xls failed. I don't know if the xls binary format (vs the xlsx XML) makes this impossible/difficult or if it just requires different methods, which I do not know.

src/XlsWorkSheet.cpp
```diff
@@ -2,51 +2,87 @@
...
+  XlsWorkBook wb = XlsWorkBook(path);
+  std::vector<ColType> types = colTypeStrings(col_types);
+  XlsWorkSheet sheet = wb.sheet(sheet_i, skip);
+  return sheet.readCols(col_names, types, na, skip);
 }
 
 // [[Rcpp::export]]
-List xls_cols(std::string path, int i, CharacterVector col_names,
-              CharacterVector col_types, std::vector<std::string> na, int nskip = 0) {
+List read_xls_(std::string path, int sheet_i, RObject col_names,
```
jennybc: This is now virtually identical to `read_xlsx_`. To the point that it feels like I should do something about that.

src/XlsWorkSheet.h
```diff
@@ -4,34 +4,38 @@
...
 #include "XlsWorkBook.h"
-#include "CellType.h"
-#include "utils.h"
+#include "XlsCell.h"
+#include "ColSpec.h"
 
 class XlsWorkSheet {
   xls::xlsWorkSheet* pWS_;
-  int nrow_, ncol_;
   double offset_;
```
jennybc: As commented earlier, it feels like this date-related information (`offset_` and `customDateFormats_`) should belong to the workbook, not the worksheet. But then I don't know how to make it available inside the member functions here, since I was not able to make the workbook a member, as is done on the xlsx side.

src/XlsWorkSheet.h
```diff
@@ -40,160 +44,276 @@ class XlsWorkSheet {
...
 
-  Rcpp::CharacterVector colNames(int nskip = 0) {
-    Rcpp::CharacterVector out(ncol_);
+  int nrow() const {
+    return nrow_;
+  }
 
-    if (nskip > nrow_)
-      return out;
+  Rcpp::CharacterVector colNames() {
```
jennybc: Almost identical to analogous member of XlsxWorksheet class. Should I do something about that?

src/XlsWorkSheet.h
```diff
@@ -40,160 +44,276 @@ class XlsWorkSheet {
...
       }
+      xcell++;
     }
-
     return out;
   }
 
-  std::vector<ColType> colTypes(const StringSet &na, int nskip = 0,
-                                 int guess_max = 1000) {
+  std::vector<ColType> colTypes(const StringSet &na,
```
jennybc: Almost identical to analogous member of XlsxWorksheet class. Should I do something about that?

src/XlsWorkSheet.h
```diff
@@ -40,160 +44,276 @@ class XlsWorkSheet {
...
     return removeSkippedColumns(cols, names, types);
   }
+
+private:
+
+  void loadCells() {
+    int nominal_ncol = pWS_->rows.lastcol;
+    int nominal_nrow = pWS_->rows.lastrow;
+
+    for (xls::WORD i = 0; i <= nominal_nrow; ++i) {
```
jennybc: The libxls goes out of its way to define and rigorously use custom types, such as `WORD`, which is `uint16_t`. readxl OTOH is all over the place with `int`, `size_t`, etc. Should I worry about this?

src/XlsWorkSheet.h
```diff
@@ -40,160 +44,276 @@ class XlsWorkSheet {
...
+
+  // Compute sheet extent (= lower right corner) directly from loaded cells.
+  //   recorded in nrow_ and ncol_
+  // Return early if there is no data. Otherwise ...
+  // Position iterators at two landmarks for reading:
+  //   firstRow_ = first cell for which declared row >= skip
+  //   secondRow_ = first cell for which declared row > that of firstRow_
+  //   fallback to cells_.end() if the above not possible
+  // Assumes loaded cells are arranged s.t. row is non-decreasing
+  void parseGeometry(int skip) {
```
jennybc: Almost identical to analogous member of XlsxWorksheet class. Should I do something about that?

src/XlsWorkSheet.cpp
hadley: Let's make that another project. Probably the right approach is to make a joint base class for XlsxWorkSheet and XlsWorkSheet etc.

src/XlsWorkSheet.h
hadley: Let's talk about that in person. It's probably somehow related to memory management.

src/XlsWorkSheet.h
```diff
@@ -40,160 +44,276 @@ class XlsWorkSheet {
...
-            REAL(col)[i] = cell.d;
-            break;
-          case COL_TEXT:
-            Rcpp::warning("Expecting numeric in [%i, %i] got `%s`",
-              i + 1, j + 1, (char*) cell.str);
-            REAL(col)[i] = NA_REAL;
-          }
+        case CELL_NUMERIC:
+        case CELL_DATE:
+          REAL(col)[row] = xcell->cell()->d;
```
hadley: I think this code would also become much more similar to the xlsx equivalent if we added `getDoubleValue()` etc methods to the cell class.

src/XlsWorkSheet.h
hadley: No

jennybc:
Yes all tests are passing everywhere and this PR added quite a few, esp for xls. So I think it is a net positive.

I'll take or create opportunities to get advice on some of the obvious opportunities for improvement, touched on above.

PR #275, opened 2017-02-26
Convert cellType() into XlsCell member function type()

jennybc:
I should have done this when I created the XlsCell class. An oversight.

Also made signature as close possible to same function for XlsxCell, in the name of convergent evolution.


jennybc:
Yes, I'm working on cell data now (types and esp. string conversion) and I think there's a lot of room to align xls and xlsx, as with geometry.

Issue #73, opened 2015-04-15
Add an option for recognizing logicals


kkmann:
Hey, I used xlsx so far, which recognizes columns containing only TRUE/FALSE as logical. Would it be possible to include some kind of support for boolean columns (maybe optional)?

jeromyanglim:
I agree. This seems almost like a bug to me. At present readxl appears to import columns with logical values in Excel as 0-1 variables.  If you have a column that contain TRUE/FALSE in Excel, then this should be imported as a character vector in R. 

Just to clarify, I would not call this an "option" as presumably it should be standard behaviour for any R function that imports Excel spreadsheets. In addition to the xlsx package, importing logical excel variables as logical in R is also standard behaviour for the gdata package.

I can imagine users assuming that TRUE/FALSE columns in Excel would be imported as logical in R. 
This would causes issues in subsequent analyses. 

It also sometimes throws the message `Unknown type: 517`.

jennybc:
Covered by #198 

@jeromyanglim If you can provide a reprex of this, please do so and open another issue:

> It also sometimes throws the message `Unknown type: 517`.

jennybc:
This will happen #270.

Issue #230, opened 2017-01-18
Report problematic cells in A1 notation


jennybc:
It would be more useful to report problems a machine readable way and, at least print, in spreadsheet coordinates, probably A1 notation (or R1C1).

``` r
library(readxl)

theURL <- "https://github.com/tts/villisika/raw/master/SpeciesObservations.xlsx"
tf <- tempfile(fileext = ".xlsx")
download.file(theURL, tf)
data_se <- read_excel(tf, sheet = "SLW Data")
#> Warning in read_xlsx_(path, sheet, col_names = col_names, col_types =
#> col_types, : [123, 70]: expecting numeric: got 'Skyddad vik vid lerig
#> älvstrand'
#> Warning in read_xlsx_(path, sheet, col_names = col_names, col_types =
#> col_types, : [139, 103]: expecting numeric: got 'Tomtskådning Dyudden'
#> Warning in read_xlsx_(path, sheet, col_names = col_names, col_types =
#> col_types, : [139, 104]: expecting numeric: got 'Arter noterade från
#> tomten Gäddeholm 2:59 på Dyuddens sydspets, med adress Gäddeholm 73, 725 97
#> Västerås. Startade 29 oktober 2006, dagen efter inflyttning. '
#> MORE WARNINGS ABOUT THE SAME COLUMNS
```

jennybc:
Suggestions from @hadley in review of #277: `cellPosition(int i, int j)` or, even better, `cellProblem(std::string message, int i, int j)`

PR #277, opened 2017-03-01
Add logical cell & col types; refactor cell typing and coercion

### LINKED ISSUE #62: `#NV` (Excel function error) translate to string `error` instead of `NA`
I read in values from a table where some cells with functions have an error and show `#NV`. While reading this data works, those function errors are translated to `error` within R. This is surprising and might lead to further errors along the way. I think, NA (the value for not defined/not available data) should be used instead.

### LINKED ISSUE #106: Force numeric when col_types specifies it
Currently we have

```
      case CELL_TEXT:
        Rcpp::warning("Expecting numeric in [%i, %i] got `%s`",
          i + 1, j + 1, (char*) cell.str);
        REAL(col)[i] = NA_REAL;
```

which fails to populate a cell with numeric data even when the user has specified a given column to be numeric.  This feature request is to parse to a numeric type in such cases.

Example:  

  u <- "https://www.spdrs.com/site-content/xls/XBI_All_Holdings.xls?fund=XBI&docname=All+Holdings&onyx_code1=1286&onyx_code2=1718"
  outfile <- tempfile(paste0(etf, administrator,"_Holdings"), fileext=".xls")
  xls.holdings <- curl::curl_download(u, outfile)
  holdings <- readxl::read_excel(outfile, skip=3, 
                                 col_types=c("text","text","numeric","text","numeric"))

### LINKED ISSUE #217: NAs for cells with 'numbers as text' in a numeric column
mre below.  thanks hadley!

	library(readxl)
	library(gdata)

	pums.layout <- paste0( tempfile() , '.xls' )

	download.file( "http://www2.census.gov/census_2000/datasets/PUMS/FivePercent/5%25_PUMS_record_layout.xls" ,	pums.layout , mode = 'wb' )

	readxl_result <- readxl::read_excel( pums.layout , sheet = 1 , skip = 1 )

	gdata_result <- gdata::read.xls( pums.layout , sheet = 1 , skip = 1 )

	# these values match what's in the file when you open in excel
	gdata_result[ 350:527 , 3 ]

	# here's the error: missings
	readxl_result[ 350:527 , 3 ]

### LINKED ISSUE #263: Dates in numeric columns
When a date is in a column that is either numeric or character, readxl silently includes the numeric representation of the date. Should this give a warning? See some discussion in #256

### LINKED ISSUE #266: Dates in Excel's internal format
I have a column which has dates in Excel's internal format. The Excel file comes from someone's reporting software, so I can't just format the column in Excel (unless I want to do it every time I get an update).

◢ |Date
-|-
1|39448
2|39449
3|39450

[date_test.xlsx](https://github.com/tidyverse/readxl/files/772728/date_test.xlsx)

These should map to the first, second, and third of January 2008.

```{r}
> as.Date(read_excel("../notes/date_test.xlsx")$Date, origin="1899-12-30")
[1] "2008-01-01" "2008-01-02" "2008-01-03"
```

But if I tell `read_excel` that it's a date: 

     z <- read_excel("../notes/date_test.xlsx", col_types="date")

```{r}
Warning messages:
1: In read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  :
  [2, 1]: expecting date: got '39448'
2: In read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  :
  [3, 1]: expecting date: got '39449'
3: In read_xlsx_(path, sheet, col_names = col_names, col_types = col_types,  :
  [4, 1]: expecting date: got '39450'
```

```{r}
> z
  Date
1 <NA>
2 <NA>
3 <NA>
```

### LINKED ISSUE #270: Support a logical column type?
It seems like an empty column should be filled with logical `NA`s but currently readxl fills with `NA_real_`. I think @hadley has already concurred, at least in principle (https://github.com/tidyverse/tidyverse/issues/34#issuecomment-278500700). Also comes up if all the cells in a column match one of the strings in `na`.

That seems easy enough to do. But it begs the question whether this should be the only way for readxl to return a logical vector. Should `"logical"` become a new column type that can be guessed or that user could specify?

Excel cells can be of Boolean type, so that can be detected when guessing or when enforcing user-specified col types.

But do you take it further? How much effort to make with logical-ish strings and numbers?

jennybc:
Main points:

  * Adds new `CellType`: CELL_LOGICAL. Detects Boolean cells, both static and formula.
  * Adds new `ColType`: COL_LOGICAL.
  * Adds these conversions:
    - logical to numeric or character
    - numeric or character to logical
    - character to numeric
    - numeric to date
  * Handle cells in error (explicit `#N/A` or as result of formula). Treat as CELL_BLANK and therefore read as `NA`.
  * Checking whether cell contents match the values in the `na` argument is now generally delegated to the `type()` member function of both XlsCell and XlsxCell. If the reported type is not CELL_BLANK, then the cell has contents you should work with. I have not exploited this to the maximum extent possible yet tbh, but am working on it.

Closes #270 Support a logical column type? (also previously discussed in #73).
Closes #263 Dates in numeric columns.
Closes #62 `#NV` (Excel function error) translates to string `error` instead of `NA`
Closes #106 Force numeric when col_types specifies it
Closes #217 NAs for cells with 'numbers as text' in a numeric column
Closes #266 Dates in Excel's internal format 
Rescues xls formula dates, even though no one has complained about that yet 🤓.

General style & housekeeping work:

  * Handle everything based on `CellType` or `ColType` enums in order.
  * Changed `return whatever; break;` into `return whatever;` inside all `switch()` statements.
  * Uniform messages/warnings across cell/col types and for xls and xlsx.
  * Make everything as similar as possible for xls and xlsx.


src/XlsCell.h
```diff
@@ -48,59 +48,73 @@ class XlsCell {
...
-        if (na.contains((char*) cell_->str)) {
+
+    case 6:    // 0x0006 formula 2.4.127 p309
+    case 1030: // 0x0406 formula (Apple Numbers Bug) via libxls
+      if (cell_->l == 0) { // formula evaluates to numeric, possibly date
+        if (na.contains(cell_->d)) {
+          return CELL_BLANK;
+        }
+        int format = styles->xf[cell_->xf].format;
+        return isDateTime(format, customDateFormats) ? CELL_DATE : CELL_NUMERIC;
```
jennybc: This rescues xls formula dates, which were previously coming in as numeric.

src/XlsCell.h
```diff
@@ -48,59 +48,73 @@ class XlsCell {
...
+      } else { // formula evaluates to Boolean, string, or error
+        if (!strcmp((char *) cell_->str, "bool")) {
+          return CELL_LOGICAL;
+        } else if (na.contains((char*) cell_->str)) {
           return CELL_BLANK;
         } else {
-          return CELL_TEXT;
+          // error should become NA
+          return cell_->d == 0 ? CELL_TEXT : CELL_BLANK;
```
jennybc: Errors come in now as `NA` instead of the string "error".

src/XlsCell.h
```diff
@@ -48,59 +48,73 @@ class XlsCell {
...
-      }
+    case 517: // 0x0205 BoolErr 2.4.24 p216:
+              //  a cell that contains either a Boolean value or an error value
+      if (na.contains((char*) cell_->str)) {
+        return CELL_BLANK;
       }
-      break;
+      // error should become NA
+      // 2.5.10 Bes p592 explains the codes, if we ever change our minds
+      return (!strcmp((char *) cell_->str, "bool")) ? CELL_LOGICAL : CELL_BLANK;
```
jennybc: Ensures explicit `#N/A` becomes `NA` instead of the string "error".

src/XlsCell.h
```diff
@@ -48,59 +48,73 @@ class XlsCell {
...
 
     default:
-      Rcpp::Rcout << "Unknown type: " << cell_->id << "\n";
-    return CELL_NUMERIC;
+      Rcpp::warning("Unrecognized cell type at [%i, %i]: '%s'",
+                    row() + 1, col() + 1, cell_->id);
     }
+
+    return CELL_TEXT;
```
jennybc: Made CELL_TEXT the default cell type vs CELL_NUMERIC. Presumably has almost zero practical significance.

src/XlsxCell.h
```diff
@@ -13,23 +14,6 @@
 // 18.3.1.96  v           (Cell Value) [p1709]
 // 18.18.11   ST_CellType (Cell Type)  [p2443]
 
-// Simple parser: does not check that order of numbers and letters is correct
```
jennybc: Moved this to utils.h because it's not necessarily specific to xlsx.

src/XlsxCell.h
```diff
@@ -54,17 +38,177 @@ class XlsxCell {
     return location_.second;
   }
 
-  std::string asStdString(const std::vector<std::string>& stringTable) const {
+  CellType type(const StringSet& na,
```
jennybc: This got moved higher in the class and substantially reworked.

src/XlsxCell.h
```diff
@@ -54,17 +38,177 @@ class XlsxCell {
...
+    //
+    // CELL_TEXT
+    //   inlineStr cell and string is found and string is not na
+    //   ISO 8601 date cell (t attr is "d") <- we're not sure this exists IRL
+    //   shared string cell and string is not na
+    //   formula string cell and string is not na
+    //   anything that is not explicitly addressed elsewhere
+  }
+
+  std::string asStdString(const StringSet& na,
```
jennybc: This was refactored to use `type()` and cover the new logical cell type.

src/XlsxWorkSheet.h
```diff
@@ -157,55 +157,133 @@ class XlsxWorkSheet {
...
+          // print date string here, when/if it's possible to do so
+          Rcpp::warning("Expecting logical in [%i, %i]: got a date",
+                        i + 1, j + 1);
+          LOGICAL(col)[row] = NA_LOGICAL;
+          break;
+        case CELL_LOGICAL:
+        case CELL_NUMERIC:
+          LOGICAL(col)[row] = xcell->asInteger(na) ? TRUE : FALSE;
           break;
         case CELL_TEXT:
```
jennybc: I am converting strings to logical as R does. At least, that is my intent.

src/XlsxWorkSheet.h
```diff
@@ -157,55 +157,133 @@ class XlsxWorkSheet {
...
-          Rcpp::warning("[%i, %i]: expecting numeric: got '%s'",
-                        row + 1, j + 1, xcell->asStdString(wb_.stringTable()));
-          REAL(col)[row] = NA_REAL;
+        {
+          std::string text_string = xcell->asStdString(na, wb_.stringTable(),
+                                                       wb_.dateStyles());
+          bool text_boolean;
+          if (logicalFromString(text_string, &text_boolean)) {
```
jennybc: I am converting strings to logical as R does. At least, that is my intent.

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,132 @@ class XlsWorkSheet {
...
+          LOGICAL(col)[row] = (xcell->cell()->d != 0) ? TRUE : FALSE;
           break;
         case CELL_TEXT:
-          Rcpp::warning("[%i, %i]: expecting numeric: got '%s'",
-                        row + 1, j + 1, (char*) xcell->cell()->str);
-          REAL(col)[row] = NA_REAL;
+        {
+          std::string text_string((char*) xcell->cell()->str);
+          bool text_boolean;
+          if (logicalFromString(text_string, &text_boolean)) {
```
jennybc: I am converting strings to logical as R does. At least, that is my intent.

src/XlsxWorkSheet.h
```diff
@@ -157,55 +157,133 @@ class XlsxWorkSheet {
...
+
+      case COL_LOGICAL:
         switch(type) {
         case CELL_BLANK:
-          REAL(col)[row] = NA_REAL;
+          LOGICAL(col)[row] = NA_LOGICAL;
           break;
-        case CELL_NUMERIC:
         case CELL_DATE:
-          REAL(col)[row] = xcell->asDouble(na);
```
jennybc: See how dates in a numeric column were treated as numeric? I don't do that anymore. They become NA.

src/XlsxWorkSheet.h
```diff
@@ -157,55 +157,133 @@ class XlsxWorkSheet {
...
+        case CELL_LOGICAL:
+          Rcpp::warning("Coercing boolean to numeric in [%i, %i]",
+                        i + 1, j + 1);
+          REAL(col)[row] = xcell->asDouble(na);
+          break;
+        case CELL_DATE:
+          // print date string here, when it's easier to do so
+          Rcpp::warning("Expecting numeric in [%i, %i]: got a date",
+                        i + 1, j + 1);
           REAL(col)[row] = NA_REAL;
```
jennybc: Dates become NA in a numeric column, instead of their integer value.

src/utils.h
```diff
@@ -1,6 +1,54 @@
...
+      col = 26 * col + (*cur - 'A' + 1);
+    } else {
+      Rcpp::stop("Invalid character '%s' in cell ref '%s'", *cur, ref);
+    }
+  }
+
+  return std::make_pair(row - 1, col - 1); // zero indexed
+}
+
+inline bool logicalFromString(std::string maybe_tf, bool *out) {
```
jennybc: Have I made this much harder than it needs to be? I borrowed the logic for looking for matches from the way the `na` argument is handled.

src/ColSpec.h
```diff
@@ -7,27 +7,59 @@
...
+  COL_LIST,    // occurs only as user-specified column type
+  COL_SKIP     // occurs only as user-specified column type
 };
 
 // ColType enum is ordered such that a simple cast
 // is sufficient to convert from CellType (user-specified
-// columns come after cell types)
-ColType inline as_ColType(CellType cell) {
-  return (ColType) cell;
+// column types come after cell types)
```
jennybc: This is mostly me moving things around.

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,132 @@ class XlsWorkSheet {
...
+        case CELL_LOGICAL:
+          Rcpp::warning("Coercing boolean to numeric in [%i, %i]",
+                        i + 1, j + 1);
+          REAL(col)[row] = xcell->cell()->d;
+          break;
+        case CELL_DATE:
+          // print date string here, when it's easier to do so
+          Rcpp::warning("Expecting numeric in [%i, %i]: got the date '%s'",
+                        i + 1, j + 1, xcell->cell()->str);
+          REAL(col)[row] = NA_REAL;
```
jennybc: Dates become `NA` in numeric columns, instead of their integer value.

src/XlsxCell.h
```diff
@@ -54,17 +38,177 @@ class XlsxCell {
...
+      rapidxml::xml_node<>* is = cell_->first_node("is");
+      std::string inline_string;
+      if (parseString(is, &inline_string)) {
+        return na.contains(inline_string) ? CELL_BLANK : CELL_TEXT;
+      } else {
+        return CELL_BLANK;
+      }
+    }
+
+    if (v == NULL || na.contains(v->value())) {
```
jennybc: This is now the main check for an empty cell or a cell with contents that match `na`. If so, the cell's type will be CELL_BLANK. In theory I could remove a lot of checks elsewhere for such a cell and just trust the cell type. But I have not done that systematically.

src/XlsxCell.h
```diff
@@ -54,17 +38,177 @@ class XlsxCell {
...
+      }
+
+      //   the mythical ISO 8601 date cell
+      //   formula string cell
+      return(v->value());
+    }
+  }
+  }
+
+  int asInteger(const StringSet& na) const {
```
jennybc: New. Needed to make logical cells.

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,149 @@ class XlsWorkSheet {
...
         case CELL_DATE:
           REAL(col)[row] = (xcell->cell()->d - offset_) * 86400;
           break;
         case CELL_NUMERIC:
-          Rcpp::warning("Expecting date in [%i, %i] got %d",
-                        i + 1, j + 1, xcell->cell()->d);
-          REAL(col)[row] = NA_REAL;
+          Rcpp::warning("Coercing numeric to date in [%i, %i]",
+                        i + 1, j + 1);
+          REAL(col)[row] = (xcell->cell()->d - offset_) * 86400;
```
jennybc: I *think* this is the right thing to do (convert numeric to date if column is `"date"`), but I'm not sure. It was requested (#266). What do you think?

src/utils.h
```diff
@@ -1,6 +1,77 @@
...
+    matches = true;
+  }
+  if (false_values.contains(maybe_tf)) {
+    *out = false;
+    matches = true;
+  }
+  return matches;
+}
+
+inline bool doubleFromString(std::string in, double& out) {
```
jennybc: I am working on a replacement for this that does not assume C++11. Travis/Appveyor have informed me that this apparently does.

jennybc: This is sorted out now.

src/XlsCell.h
```diff
@@ -48,59 +48,73 @@ class XlsCell {
...
+    case 6:    // 0x0006 formula 2.4.127 p309
+    case 1030: // 0x0406 formula (Apple Numbers Bug) via libxls
+      if (cell_->l == 0) { // formula evaluates to numeric, possibly date
+        if (na.contains(cell_->d)) {
+          return CELL_BLANK;
+        }
+        int format = styles->xf[cell_->xf].format;
+        return isDateTime(format, customDateFormats) ? CELL_DATE : CELL_NUMERIC;
+      } else { // formula evaluates to Boolean, string, or error
+        if (!strcmp((char *) cell_->str, "bool")) {
```
hadley: I think you should use `strncmp()` here to be safe (in case `cell_->str` doesn't contain a `\0`)

src/XlsCell.h
hadley: Why is that implied by `cell_->d == 0`? Might be worth a comment.

src/XlsCell.h
hadley: `strncmp()` here too. Maybe extract out into helper function?

src/XlsWorkSheet.h
```diff
@@ -81,9 +81,7 @@ class XlsWorkSheet {
 
     // no cell data to consult re: types
     if (xcell == cells_.end()) {
-      for (size_t i = 0; i < types.size(); i++) {
-        types[i] = COL_BLANK;
-      }
+      std::fill(types.begin(), types.end(), COL_BLANK);
```
hadley: 👍

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,149 @@ class XlsWorkSheet {
...
         switch(type) {
         case CELL_BLANK:
-          REAL(col)[row] = NA_REAL;
+          LOGICAL(col)[row] = NA_LOGICAL;
           break;
-        case CELL_NUMERIC:
         case CELL_DATE:
-          REAL(col)[row] = xcell->cell()->d;
+          // print date string here, when/if it's possible to do so
+          Rcpp::warning("Expecting logical in [%i, %i] got a date",
```
hadley: I wonder if it's worth pulling out a `cellPosition(int i, int j)` now? That would make it easier to implement #230. (But this PR is already quite big, so feel to leave)

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,149 @@ class XlsWorkSheet {
...
+                        i + 1, j + 1);
+          LOGICAL(col)[row] = NA_LOGICAL;
+          break;
+        case CELL_LOGICAL:
+          LOGICAL(col)[row] = (xcell->cell()->d == 1) ? TRUE : FALSE;
```
hadley: `(xcell->cell()->d == 1)` should be adequate

src/XlsWorkSheet.h
hadley: That seems consistent with the new behaviour in the other branches (i.e. try harder to be helpful).  I think you could drop the warning.

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,149 @@ class XlsWorkSheet {
...
         }
         break;
+
+      case COL_NUMERIC:
+        switch(type) {
+        case CELL_BLANK:
+          REAL(col)[row] = NA_REAL;
+          break;
+        case CELL_LOGICAL:
+          Rcpp::warning("Coercing boolean to numeric in [%i, %i]",
```
hadley: In general, I don't think you need a warning if the user has explicitly asked for the coercion.

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,149 @@ class XlsWorkSheet {
...
+                        i + 1, j + 1);
+          REAL(col)[row] = xcell->cell()->d;
+          break;
+        case CELL_DATE:
+          // print date string here, when/if possible
+          Rcpp::warning("Expecting numeric in [%i, %i]: got a date",
+                        i + 1, j + 1);
+          REAL(col)[row] = NA_REAL;
```
hadley: Shouldn't this be `xcell->cell()->d` ?

src/XlsWorkSheet.h
```diff
@@ -146,61 +144,149 @@ class XlsWorkSheet {
...
       case COL_LIST:
         switch(type) {
         case CELL_BLANK: {
           SET_VECTOR_ELT(col, row, Rf_ScalarLogical(NA_LOGICAL));
           break;
         }
-        case CELL_NUMERIC: {
-          SET_VECTOR_ELT(col, row, Rf_ScalarReal(xcell->cell()->d));
+        case CELL_LOGICAL: {
+          SET_VECTOR_ELT(col, row, Rf_ScalarLogical((xcell->cell()->d)));
```
hadley: Two many parens?

src/XlsxCell.h
hadley: I think that would be a good idea.

src/XlsxWorkSheet.h
```diff
@@ -269,7 +362,7 @@ class XlsxWorkSheet {
     ncol_ = 0;
 
     // empty sheet case
-    if (cells_.size() == 0) {
+    if (cells_.empty()) {
```
hadley: This seems like a lot of duplicated code. I think it's probably worth the effort to define a common base class for `XlsxCell` and `XlsCell` so we only need one function instead of two. That can happen after this PR though.

src/utils.h
hadley: I think you probably should initialise `true_strings` and `false_strings` once, outside of this function. @jimhester can suggest the best way

src/XlsWorkSheet.h
jennybc: Unfortunately, we don't actually know in this function whether the column type was set by the user or via guessing.

src/utils.h
jimhester: I believe R exposes Rf_StringTrue in `R_ext/Utils.h`, is it possible we could use that rather than rolling our own here.

If that is not possible I would do something like https://github.com/jennybc/readxl/pull/2

hadley: Yeah looks like Rf_StringTrue is the way to go

src/XlsCell.h
jennybc: done cf46fcfbdbfefbfca497f442932f1225126e752e

src/XlsCell.h
jennybc: done 699a5c8d3f352c6caf51c2efa6fc2f60e078a3bb

src/XlsCell.h
jennybc: done cf46fcfbdbfefbfca497f442932f1225126e752e (well, not the helper function part (yet?))

src/XlsWorkSheet.h
jennybc: done 5edc343ebb576ac2e7c8af51a44538da9f8820e3

src/XlsWorkSheet.h
jennybc: At the moment, the choice to not coerce a date to numeric is intentional and consistent between xls and xlsx. See #263. Does anyone ever want that integer mixed in with other numbers that are not dates? Yet I can see how this seems in conflict with all the other places where we've added coercion.

src/XlsxCell.h
jennybc: Still need to do this!

src/XlsWorkSheet.h
jennybc: done bd4eacfd29403045d45ab4d5a760c8cfc0f78828

src/XlsxCell.h
jennybc: Maybe cell type should become part of the Xls[x]Cell class and determined + stored when the cell is loaded. Otherwise, we re-type many cells up to 3 times: col type guessing, insertion of cells into columns, and coercion from one type to another.

src/utils.h
jennybc: done adc8151d9fdb6e23d12cee4c49f62ad2a760f271

src/XlsWorkSheet.h
hadley: And actually it would probably be better to have a `cellProblem(std::string message, int i, int j)` but that definitely doesn't belong here

src/XlsWorkSheet.h
hadley: Oh I see. This will become less annoying once we have readr style problems tibble.

src/XlsWorkSheet.h
hadley: Yeah, it seems like reasonable behaviour but not in keeping with the other places where we do conversion. It might be worth spending a couple of minutes trying to articulate the principle of why this is ok here.

src/XlsxCell.h
hadley: Yeah, that seems like a good idea. Maybe do as a next step in a separate PR?

src/XlsxCell.h
```diff
@@ -54,17 +38,177 @@ class XlsxCell {
...
+
+    Rcpp::warning("Unrecognized cell type at [%i, %i]: '%s'",
+                  row() + 1, col() + 1, t->value());
+
+    return CELL_TEXT;
+
+    // summary of how Excel cell types have been mapped to our CellType
+    //
+    // CELL_BLANK
+    //   inlineStr cell and (string is na or string can't be found)
```
hadley: I feel like this might be more informative at the start of the function.

src/XlsxCell.h
```diff
@@ -54,17 +38,177 @@ class XlsxCell {
...
-    int id = atoi(v->value());
-    return stringTable.at(id);
+    case CELL_TEXT:
+    {
+      std::string out_string;
+
+      // inlineStr
+      rapidxml::xml_node<>* is = cell_->first_node("is");
+      if (is != NULL) {
+        return parseString(is, &out_string) ? out_string : "NA";
```
hadley: It's a bit odd to see a literal `"NA"` here

src/utils.h
```diff
@@ -1,6 +1,61 @@
...
+    *out = true;
+    matches = true;
+  } else if (Rf_StringFalse(maybe_tf.c_str())) {
+    *out = false;
+    matches = true;
+  }
+  return matches;
+}
+
+inline bool doubleFromString(std::string mystring, double& out) {
```
hadley: It might be a bit more idiomatic to return a `std::pair<bool, double>` here. 

What do you think @jimhester?

src/XlsWorkSheet.h
jennybc: I've put a note in #198 about tracking whether a column's type was explicitly specified vs. guessed. Then the warning could be suppressed if the user specified the column type.

I will make a table of the types and coercions and try to say something principled about which ones work, which ones warn, etc. And if I can't, will make it so!

src/XlsxCell.h
jennybc: Issue opened #283

src/XlsxWorkSheet.h
jennybc: Issue opened #284

src/utils.h
jennybc: It feels like the various solutions I read on SO work this way. I also note it is how the existing `parseString()` works: https://github.com/tidyverse/readxl/blob/f95e0ae1d963c87dd7fd66c876116e84ae161657/src/XlsxString.h#L33

hadley: Yeah, it's a totally reasonable pattern, but for some reason it feels off to me here.

src/XlsxCell.h
jennybc: OK done in a9402dcf137d8b201b6318dfea036e9707c5aac2

hadley:
I think this is fine to merge if you want to; other stuff can happen later since this is already a gigantic PR.

jennybc:
That would be a great mercy (merging and attacking to do's in smaller pieces). Will do that.

PR #285, opened 2017-03-06
xlsx: make cell type a data member; see #283

jennybc:
Eliminate repeated re-typing and widespread downstream comparisons to the NA strings, in favor of doing once when the cell is loaded.

One downside to this is that we are now typing all cells, even in columns of type `"skip"` and in rows destined to be skipped. Could we add cell type as a data member but not populate it in the constructor? Or set it to a placeholder value, e.g. `CELL_UNKNOWN`. And then determine & store the cell type the first time we actually need it?

*I'm doing xlsx and xls separately, because I'm concerned xls will be harder.*

### DIFF
--- a/src/XlsxCell.h
+++ b/src/XlsxCell.h
@@ -17,17 +17,25 @@
 class XlsxCell {
   rapidxml::xml_node<>* cell_;
   std::pair<int,int> location_;
+  CellType type_;
 
 public:
 
   // if possible, provide guess at row and column based on position within xml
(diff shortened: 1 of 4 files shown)


hadley:
I think it would make sense to compute on demand. To do that add a `CELL_UNKNOWN` type, and then in `type()` if the member field is set to unknown, compute and save it.  I'm not sure how to make sure that the styles and strings will be available at the right time.

You'll need to use the mutable type specifier on the type field so that the type accessor function can still be const (because the while data in the object is changing the external interface is not).

jennybc:
Closing this. Will wait until #286 gets merged and resume this task based on that.

PR #290, opened 2017-03-11
Date subsecond rounding; fixes #187

### LINKED ISSUE #187: Date/times missing a second
Not sure if this is an error but if in Excel you enter `4/28/2016  11:30:00 AM` and read the file using readxl [for example, `readxl::read_excel("./test.xlsx",col_names = FALSE)`] you end up losing a second. Also the package assumes is UTC. See related post at http://stackoverflow.com/questions/38005379/excel-date-time-imported-into-r-missing-1-second

jennybc:
I know this is sort of horrible. But it's low tech and seems to work. The test sheets currently contain two cells: the problematic date-time from #187 and another from the analogous issue thread from the SheetJS world. Before this fix, the test fails and after it passes. My intention is to produce a date-time in R that matches the one in Excel to at least the second. I recommend you stop reading here.

*Notes after excessive amounts of study*: You might think this problem only depends on time of day, since Excel's serial dates use a natural unit of 1, with the decimal part capturing the time of day. And you might then create a comprehensive test sheet with one row per second in a day and use various ways to test that you import this sheet correctly, at least to 1 second accuracy. But you would be wrong.

The floating point number recorded by Excel for a specific time of day depends, in completely mysterious ways, on both the day and whether the date-time is hard-coded ("typed") vs. formula vs. "paste by value" from a formula. For example, the problematic date-times mentioned above are NOT problematic in the context of a sheet that holds one row per second in the day or for other days in history.

Here are the numbers recorded in `xl/worksheets/sheet???.xml` for a cell holding 11:30am in various scenarios:

  * Hand-typed for 2016-04-28: 42488.479166666664
  * Computed via formula, by adding 1 second repeatedly to a base of 00:00:00 on 2016-04-28: 42488.479166778678
  * Hand-typed, same time of day, but on 2017-01-01: 42736.479166666664
  * Hand-typed, same time of day, but on 1954-01-01: 19725.479166666668

I think you can only trust Excel's floating point representation of a date-time to the n-th decimal place. You should go part way there, round, send that info to R, and be thankful you got this far.

Another exercise that yielded super-weird results: I write one date-time for each second in a specific day. Import that into Excel and write to xlsx. Import that xlsx into R and take the sequential differences via `diff.POSIXt()`. They should all be 1. But they are not. And the actual values, cardinality, etc. vary in impenetrable ways, depending on which day in history you examine.


jennybc:
You missed my epic account of today's adventures. Wow.

Issue #148, opened 2015-12-01
does convert dates correctly in the ranges of 1/1/1900 to 2/27/1900


wjcgh:
Because Excel replicated the Lotus 1-2-3 2/29/1900 bug, there is an off-by-one day conversion error for dates in the ranges of 1/1/1900 to 2/28/1900, which are converted to 12/31/1899 to 2/27/1900. I attached a test file for inclusion in your date test case.
[testdate.xlsx](https://github.com/hadley/readxl/files/48883/testdate.xlsx)

jennybc:
Duplicate of #264

PR #292, opened 2017-03-13
Account for leap year bug; fixes #264

### LINKED ISSUE #264: Difference of one day.
Hello, 

I create a `xlsx` file with a "Time" formatting:
```
library(openxlsx)
dat <- data.frame(A=c(1, 1.5, 1.8))
wb <- createWorkbook()
addWorksheet(wb=wb, sheetName = "Sheet1")
sty <- createStyle(numFmt = "h:mm AM/PM")
addStyle(wb, sheet=1, style=sty, rows=2:4, cols=1)
writeData(wb, sheet=1, dat)
saveWorkbook(wb, "test02.xlsx", overwrite = TRUE) 
```

In Excel:

capture

But:

```
> readxl::read_excel("test02.xlsx", sheet=1)
                    A
1 1899-12-31 00:00:00
2 1899-12-31 12:00:00
3 1899-12-31 19:12:00
```

jennybc:
Excel reproduces a bug from Lotus 1-2-3: the non-existent leap day February 29, 1900 is included in the date system. This only affects sheets that use the 1900 date system and dates prior to March 1, 1900.

The proposed fix effectively removes that date. I adjust (increment) date-times on or before February 28, 1900. I convert date-times on February 29 1900 -- a day which does not actually exist -- to NA and throw a warning.

Since knowledge of the date system is now required for both the leap year adjustment and the offset, I store and pass the 1900 vs 1904 indicator around instead of the offset.

Fixes #264 and #148 (already closed as duplicate), if you want to read more about it.


src/utils.h
```diff
@@ -29,6 +50,35 @@ inline double dateRound(double dttm) {
...
+// this is even more horrible
+// correct for Excel's faithful re-implementation of the Lotus 1-2-3 bug,
+// in which February 29, 1900 is included in the date system, even though 1900
+// was not actually a leap year
+// https://support.microsoft.com/en-us/help/214326/excel-incorrectly-assumes-that-the-year-1900-is-a-leap-year
+// How we address this:
+// If date is *prior* to the non-existent leap day: add a day
+// If date is on the non-existent leap day: make negative and, in due course, NA
+// Otherwise: do nothing
+inline double removeLeapDay(double xlDate) {
```
hadley: I think this would be slightly easier to follow if you combined these two functions.

jennybc: Yes good point. Done 2d57d7a389a672e46090f45881bf9692b322fd0b

wjcgh:
Thank you jennybc for making the change.

PR #303, opened 2017-03-25
Exclude vendored code from test coverage

jennybc:
I want to see the coverage report for this.


jennybc:
Merging. I might rework `.travis.yml` at some point, but just for aesthetic reasons.

cc @hadley

Issue #284, opened 2017-03-05
Unified cell/worksheet classes or generic functions


jennybc:
so that all the code now duplicated in xls & xlsx can, instead, be shared.

PR #304, opened 2017-03-26
Refactor and align date format handling

jennybc:
Main point is to prepare for greater unification of xls and xlsx processing (#284).

This makes the date format handling on the xls side match that from xlsx as closely as possible.

Since I had to read all the docs on numeric/date formatting, I made some other superficial changes:

  * "style(s)" --> "format(s)". Because that's really what we mean and using that vocabulary makes it less confusing to read the docs.
  * Updated my beloved copy of ECMA-376 Part 1 to the 5th edition and (hopefully) all the associated page numbers.
  * Moved some more notes into comments for future re-orientation.

### DIFF
--- a/src/XlsxWorkSheet.h
+++ b/src/XlsxWorkSheet.h
@@ -8,18 +8,23 @@
 #include "ColSpec.h"
 
 // Page and section numbers below refer to
-// ECMA-376
-// version, date, and download URL given in XlsxCell.h
+// ECMA-376 (version, date, and download URL given in XlsxCell.h)
 // 18.3.1.73  row         (Row)        [p1685]
 // 18.3.1.4   c           (Cell)       [p1593]
 // 18.3.1.96  v           (Cell Value) [p1707]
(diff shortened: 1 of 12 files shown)


src/XlsxWorkSheet.h
```diff
@@ -7,13 +7,13 @@
...
 #include "ColSpec.h"
 
-// Key reference for understanding the structure of the XML is
-// ECMA-376 (http://www.ecma-international.org/publications/standards/Ecma-376.htm)
-// Section and page numbers below refer to the 4th edition
-// 18.3.1.73  row         (Row)        [p1677]
-// 18.3.1.4   c           (Cell)       [p1598]
-// 18.3.1.96  v           (Cell Value) [p1709]
-// 18.18.11   ST_CellType (Cell Type)  [p2443]
+// Page and section numbers below refer to
```
hadley: The line breaking here is a little odd

jennybc: Fixed that and, more importantly, one more small thing that was different. Merging.

Issue #309, opened 2017-03-30
read_excel importing NAs when importing date and time from third-party xls file


ea-guerette:
I am experiencing an issue with read_excel, reading xls files from a third-party. 

I put a cropped example file on Cloudstor as I cannot attach it here: 
https://cloudstor.aarnet.edu.au/plus/index.php/s/EK9NViho5MaxIGF

When I read the file using:

df <- read_excel("reprex_file_readxl.xls", sheet = 1, skip = 3, col_names = F, na = c("", "--"),
                 col_types = c("text", "text", "numeric", "numeric")) 

The first two columns are NAs. I tried setting the col_types to "date", and to "guess", but it did not work either. 

I downloaded readxl from git_hub using:
devtools::install_github("tidyverse/readxl")
(the CRAN version worked fine for the first two columns, but would drop the last column (which is a known issue (#99)).

> sessionInfo()
R version 3.3.1 (2016-06-21)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] readxl_0.1.1.9000

loaded via a namespace (and not attached):
[1] clipr_0.3.2    assertthat_0.1 tools_3.3.1    tibble_1.2     Rcpp_0.12.10   knitr_1.15.1

jennybc:
I see the problem and can fix it. Two questions:

  1. Can I use the example you posted as a test sheet in the package? Obviously it represents an edge case and, once I fix it, this will help make sure it stays fixed.
  2. Do you know the name of the software that made this file? I like to keep track of such things.

ea-guerette:
Hi,

Thank you!


1. Yes, you can use it as a test sheet.

2. I do not know, but I can probably find out.


Cheers,


Elise

________________________________

jennybc:
@ea-guerette You should be able to read this file if you install from a pull request I just made:

```r
devtools::install_github("tidyverse/readxl#312")
```

Something along these lines will be merged into master before too long, but this should work for now. I'd be interested to know either way, in fact.

ea-guerette:
Hi,

It worked!

[...]

PR #293, opened 2017-03-19
Use uint32_t to index into xls string table; fixes #129

jennybc:
libxls clearly anticipates that the string table size might require an unsigned 32 bit integer, e.g.:

https://github.com/tidyverse/readxl/blob/0587f00dcc889bc263b232da4c1951c631b975f3/src/libxls/xlsstruct.h#L240-L246

https://github.com/tidyverse/readxl/blob/0587f00dcc889bc263b232da4c1951c631b975f3/src/libxls/xlsstruct.h#L398-L412

https://github.com/tidyverse/readxl/blob/0587f00dcc889bc263b232da4c1951c631b975f3/src/xls.c#L137

https://github.com/tidyverse/readxl/blob/0587f00dcc889bc263b232da4c1951c631b975f3/src/xls.c#L152

(all these types are defined [here](https://github.com/tidyverse/readxl/blob/0587f00dcc889bc263b232da4c1951c631b975f3/src/libxls/xlstypes.h#L39-L47))

But then it was indexing into the table with a 16 bit integer and therefore vulnerable to getting the wrong strings. It takes a large sheet to expose this, i.e. more than 65535 unique strings. Which is the case with the xls from #129.

Two questions:

  * Should I add a test for this? It would require a very large xls, plus it doesn't really seem like something readxl should be responsible for.
  * Should I try to push this change back into libxls? It's not trivial for me, because it's on sourceforge and uses svn, whereas I've only done this with Git/GitHub. Yet it's a tiny change and seems like a no-brainer.

PR #310, opened 2017-03-30
LABEL records in xls 1) exist and 2) have length stored as 16-bit values; fix…

jennybc:
…es #309 

In #309, we have an xls BIFF5, a very old format (but with recent data, so apparently some tool is still writing this!). It uses the LABEL record type to hold strings, which is an ancient practice and, I thought, no longer occurring. I was wrong.

Problem 1: readxl had stopped reading LABEL cells. You had that type there, but I'd let it go missing in one place. I've added it back and such cells are now loaded again.

Problem 2: The fix re: using longer integers to index into the shared string table (https://github.com/tidyverse/readxl/pull/293) actually breaks the string extraction for LABEL records. All of this is libxls problems/bugs btw. The problem is that `label` needs to be a pointer to `uint32_t` for parsing modern BIFF8 sheets / LABELSST records and a pointer to `uint16_t` for sheets like the one in #309.

I don't think is the right solution yet, but need advice. It does fix the problem and tests pass, but memcheck shows many instances of "invalid read", "invalid write", "Address 0x10fbe338c is 4 bytes after a block of size 8 alloc'd", "Conditional jump or move depends on uninitialised value(s)", "Use of uninitialised value of size 8".

I suppose I could contain the entire fix inside the `if(pWB->is5ver) {...}`, which would limit its execution to the very rare scenario from #309. Ideas?


hadley:
I think @jimhester is probably better for this review than me

jennybc:
@jimhester I'd love your advice re: how to do this while keeping memcheck happy.

Here's the info I have on the LABEL record and why I fixed this the way I did. The sheet in #309 is BIFF5.

p187 of
http://www.openoffice.org/sc/excelfileformat.pdf

*The more definitive and modern reference on the xls seems to describe BIFF8 only and not the BIFF5 format we have here. 😔*

`label` in our case is a pointer to this variable length "Byte string, 16-bit string length"

screen shot 2017-03-30 at 7 26 07 am

p17 of
http://www.openoffice.org/sc/excelfileformat.pdf

Explains BIFF2-BIFF5 byte strings.

screen shot 2017-03-30 at 7 36 51 am

jennybc:
Closing this in favour of #312.

PR #320, opened 2017-04-05
Completion support for readxl_example

jimhester:
This uses utils:::.addFunctionInfo to provide additional completions for
specific functions. Because this is completion for argument names rather
than string values we need to modify readxl_example to take unquoted
names for it to be useful.

To try it out load the package then type `readxl_example(<TAB>)`

This is a proof of concept, I am not sure this is actually a good idea.


jimhester:
https://github.com/tidyverse/readxl/pull/320/commits/c595dc5740e0af816d05f20d4ab4bf8e7e688ccb defines a custom completer instead of the previous approach, I think it is more what you were looking for. It works pretty well apart from some fiddlyness with detecting the function.

jennybc:
I've tried it out in RStudio and the shell and it's a cool concept!

Right now it's completing the full, relative path name which, at least in RStudio, is default behaviour. And then, because of the way I've written `readxl_example()`, that path can't be found, due to the leading `"inst/extdata/"`.

``` r
library(readxl)
readxl_example(path = "inst/extdata/datasets.xls")
#> Error in system.file("extdata", path, package = "readxl", mustWork = TRUE) : 
#>   no file found
readxl_example(path = "datasets.xls")
#' [1] "/Users/jenny/resources/R/library/readxl/extdata/datasets.xls"
```

jimhester:
> Right now it's completing the full, relative path name

If it's doing that than RStudio's filename completions are overriding this, try it in a regular R console.

hadley:
cc @kevinushey

kevinushey:
Right now, RStudio doesn't respect the `custom.completer` option -- I'll file an issue on the RStudio side to add support for that.

songguoer22:
Hello，why  excel_sheet funtion  can  only read some xlsx file in the readxl_example?
This is my code and some problem,please give me some help,Thanks.
"> library(readxl)
> setwd("E:/R/readxl/")
> city_code <- read_excel("city_code.xlsx")
> excel_sheets(city_code)
Error in file.exists(path) : invalid 'file' argument
"

hadley:
@songguoer22 please file a new issue with a reproducible example. I am deleting your comment since it is not relevant to this pull request

jimhester:
@jennybc I just redid this PR to use the new [completeme](https://github.com/jimhester/completeme) package, which supports multiple custom completions from different packages.

screenshot 2017-08-23 09 59 25

Also thanks to @kevinushey's https://github.com/rstudio/rstudio/pull/1114 user defined completions now work in recent version fo the IDE, so this works there as well.

screenshot 2017-08-23 09 58 30

jennybc:
This looks very very awesome! In general and here. 😄

jennybc:
Is this ready to merge or should I let it simmer a bit longer?

What's the approximate timeframe for completeme going to CRAN?

jimhester:
Not sure on either question, it can probably simmer a bit longer

PR #326, opened 2017-04-07
Trim leading and trailing whitespace, by default; closes #211

### LINKED ISSUE #211: feature request: trim whitespace on read
It'd be great to have an option like strip.white=TRUE, or strip_white=BOTH to automatically remove leading or trailing whitespace on import.

jennybc:
This is the same PR you reviewed and approved already (#325), plus one commit. Why a new PR? Minor Git snafu.

First commit is the same.

Second commit deals with this: if `trim_ws = TRUE` then whitespace should be trimmed prior to any comparison with the `na` strings.


jennybc:
Also, you're going to see more test sheets (re-)named like so: `tests/testthat/sheets/whitespace-xlsx.xlsx`. It looks silly, but when you have the workbook open in Excel it is damn near impossible to tell if you're working on the xls or xlsx version. Repeating the extension in the filename makes it obvious.

jennybc:
BTW this matches readr's behaviour re: `na` and `trim_ws`: trim, then check for NAs.

Issue #241, opened 2017-01-31
Bring readxl into the tidyverse


jennybc:
### Initial transfer

  * [x] Move to tidyverse organisation.
  * [x] Search for `hadley/` and replace with `tidyverse/`.
  * [x] Modify `.git/config`

### travis

  * [x] Make sure includes:

```yaml
r:
 - oldrel
 - release
 - devel
```

### pkgdown

  * [x] Add template `_pkgdown.yml`

```yaml
template:
  package: tidytemplate
  default_assets: false
```

  * [x] Add to `.Rbuildignore`

  * [x] `build_site()` then push.

  * [x] In admin use `docs/` for documentation and turn off wiki

  * [x] create `docs/CNAME` consisting of `PACKAGE.tidyverse.org`

  * [x] Update url on github repo page.

  * [x] Add url to DESCRIPTION.

  * [x] Group functions deliberately in function reference.

### Logo

  * [x] Copy in hex logo (120 x 139) ~~and add to `.Rbuildignore`~~.

  * [x] Package name + hex logo.

```
# {{package name}} <img src="logo.png" align="right" />
```

  * [x] Add image + link on <http://tidyverse.org>

### `README.Rmd`

  * [x] Switch from `.md` to `.Rmd`

  * [x] Install instructions:

````r
```{r, eval = FALSE}
# The easiest way to get readr is to install the whole tidyverse:
install.packages("tidyverse")

# Alternatively, install just readr:
 install.packages("readr")

# Or the the development version from GitHub:
# install.packages("devtools")
devtools::install_github("tidyverse/readr")
```
````

  * [x] Overview: 1-2 paragraphs including brief description/goals. Add h2, and set:

   ```yaml
   home:
     strip_header: true
   ```

  * [x]  Link to appropriate R4DS chapter:

   ```md
   If you are new to readr, the best place to start is the
   [data import chapter](http://r4ds.had.co.nz/data-import.html) in R
   for data science.
   ```

  * [x] Add link to `_pkgdown.yaml`

   ```yaml
     links:
     - text: Learn more
       href: http://r4ds.had.co.nz/data-import.html
   ```

  * [x]  Usage section shows a few examples.

### Package help

  * [x] Review package title and description in DESCRIPTION.

  * [x] Ensure that RStudio is listed as funder.

  * [x] Add package help page.

   ```r
   #' @keywords internal
   "_PACKAGE"
   ```

  * [x] Switch to markdown, while there is still so little documentation *JB addition*

[...]

PR #328, opened 2017-04-10
Init 'workflows' article; closes #323, relates to #241

### LINKED ISSUE #323: Document writing to csv immediately after `read_excel()`
Re: reproducibility.

One of the reasons I think packages like readxl are important is because they allow reading of data as you found it ... even if that's in xls or xlsx.

But I also like to make my workflows transparent, future proof, and version controlled. So I usually save spreadsheet data to plain text form at the same time as I read it into R, just to have a non-binary, plain text snapshot of what the sheet looked like at that moment.

Should `read_excel()` have an argument that applies `readr::write_delim()` to the tibble its returning? Or should I just put sample user code in a vignette or article about using readxl in workflows? The example currently in `excel_sheets()` about [reading all the worksheets at once](http://readxl.tidyverse.org/reference/excel_sheets.html) is another good candidate. And the combination would be great: snapshot a multi-worksheet workbook in a series of csv files.

cc @hadley

jennybc:
Brain dump on "readxl workflows". I know I still need to create a more basic vignette (#241).

This shows how to cache to CSV at same time as read (#323), as well as some `lapply()`/`purrr::map_*()` workflows for iterating over worksheets.

2 questions:

  * Do you even want to review docs/vignettes? BTW rendered version of this is [here](http://rpubs.com/jennybc/readxl-workflows).
  * When to use a vignette vs an article? Main reason I ask is re: dependencies. This vignette could possibly benefit from `library(tidyverse)`, but I'm not sure I even want to Suggest the entire tidyverse.


hadley:
I'm happy to review anything that you'd like feedback on. Here, I'd have a mild preference for `dir()` and `map()` over `list.files()` and `lapply()`. I'll note a couple of other minor things too.

I think CRAN convenience is one reason to prefer articles over vignettes. I'm not sure if people will discover articles as easily as vignettes. Time will tell.

vignettes/readxl-workflows.Rmd
```diff
@@ -0,0 +1,180 @@
...
+delete_on_exit <- character()
+```
+
+*under development*
+
+Ideas for using readxl to increase reproducibiliy and reduce tedium.
+
+```{r}
+library(readxl)
+`%>%` <- magrittr::`%>%`
```
hadley: Maybe readxl should re-export the pipe?

vignettes/readxl-workflows.Rmd
```diff
@@ -0,0 +1,180 @@
...
+If raw data is stored *only* as an Excel spreadsheet, this limits your ability to inspect it when solving the little puzzles that crop up in dowstream work. You'll need to fire up Excel (or similar) and get busy with your mouse. You certainly can't poke around it or view diffs on GitHub.
+
+Solution: cache a CSV snapshot of your raw data tables at the time of export. Even if you use `read_excel()` for end-to-end reproducibility, this complementary CSV leaves your analysis in a more accessible state.
+
+Pipe the output of `read_excel()` directly into `readr::write_csv()` like so:
+
+```{r}
+library(readr)
+
+iris_xl <- read_excel(readxl_example("datasets.xlsx"), sheet = "iris") %>%
```
hadley: I think it would be slightly better to write:

```R
iris_xl <- readxl_example("datasets.xlsx") %>%
  read_excel(sheet = "iris") %>% 
  write_csv("iris-raw.csv")
```

since that will more closely parallel the multi-file pipe.

vignettes/readxl-workflows.Rmd
```diff
@@ -0,0 +1,180 @@
...
+Even though the worksheets in `deaths.xlsx` have the same layout, we'll specify the target rectangle in two different ways here, to demonstrate how this works if each worksheet has it's own peculiar geometry.
+
+```{r}
+path <- readxl_example("deaths.xlsx")
+(sheets <- path %>%
+    excel_sheets() %>% 
+    set_names())
+ranges <- list("A5:E15", cell_rows(5:15))
+my_fun <- function(x, path, range) {
+  write_csv(read_excel(path = path, sheet = x, range = range),
```
hadley: Would you really save the individual sheets here and not the one giant file?

vignettes/readxl-workflows.Rmd
```diff
@@ -0,0 +1,180 @@
...
+xl_list <- lapply(excel_sheets(path), read_excel, path = path)
+xl_list
+```
+
+### CSV caching and iterating over sheets
+
+What if we want to read all the sheets in at once and simultaneously cache to CSV? We define `my_fun()` as `write_csv(read_excel(...))` and use `lapply()` again. We could even define this on-the-fly as an anonymous function inside `lapply()`, but I think this is more readable.
+
+```{r}
+my_fun <- function(x, path) {
```
hadley: I think `my_fun()` deserves a better name.

Just for discussion, here's a more pipe-y version:

```R
path %>%
  excel_sheets() %>%
  map(. %>% 
    read_excel(path, sheet = .) %>% 
    write_csv(paste0("datasets-", x, ".csv")
  )
```

vignettes/readxl-workflows.Rmd
jennybc: I think I'm going to make this an article, fully embrace the `verse from the start, and include base code in an Appendix or something. So the pipe will be available.

vignettes/readxl-workflows.Rmd
jennybc: Luckily we don't have to worry about this. Although I defined such a function, I promptly forgot to use it inside the map. 😁 Will revise on next pass.

vignettes/readxl-workflows.Rmd
jennybc: done in e74ae1d

vignettes/readxl-workflows.Rmd
jennybc: As of 7154439, this function is `read_then_csv()`.

jennybc:
I am content with the current state, which you can see rendered here: http://rpubs.com/jennybc/readxl-workflows.

This is no longer a vignette, so the "real" version will be the pkgdown version, which can be updated at any time. So there is no sense in which we're stuck with the version at the time of CRAN submission.

As I write this, though, I'm having a hard time getting travis to gracefully accept the presence of a non-vignette article. Appveyor does not seem to mind. I'll have to check the outcome in the morning.

jennybc:
I think the trouble was initiating a vignette, then downgrading it to an article, with no genuine vignette remaining. Now that I've removed `VignetteBuilder` from DESCRIPTION and Rbuildignored `vignettes/articles`, the build passes.

PR #335, opened 2017-04-14
Write sheet geometry vignette

jennybc:
It's not exactly a page turner, but this needed to be written somewhere.


jennybc:
http://rpubs.com/jennybc/267885

jennybc:
I'm just going to merge this. But the PR structure should still make it easy for you provide comments, if you feel so moved.

PR #337, opened 2017-04-15
Registration of native routines; fixes #331

### LINKED ISSUE #331: register compiled functions
For any flavour that just has a NOTE on readxl's CRAN check results? This is what it's about. So this will clear all of those.

https://cran.rstudio.com/web/checks/check_results_readxl.html

jennybc:
I can't check against rdevel on rhub atm because it appears to be down. But travis includes an rdevel run, so it will be interesting to confirm that the associated NOTE goes away.


jennybc:
Yes travis checks confirm the NOTE has gone away.

PR #341, opened 2017-04-17
Stop guessing if we're already at COL_TEXT

jennybc:
This is a minor tweak motivated by speed. Example: for one single large xlsx, created from a single wild-caught xls (>50K rows, >100 cols), this speeds up `read_excel()` by >25%. The speed up for xls version of this sheet is closer to 50%. Highly anecdotal, but seems like a sensible change regardless.

### DIFF
--- a/src/XlsWorkSheet.h
+++ b/src/XlsWorkSheet.h
@@ -112,7 +112,7 @@ public:
         Rcpp::checkUserInterrupt();
       }
       int j = xcell->col() - actual_.minCol();
-      if (type_known[j]) {
+      if (type_known[j] || types[j] == COL_TEXT) {
         xcell++;
         continue;
       }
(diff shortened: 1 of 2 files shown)


jennybc:
FYI @hadley but I'm just merging this.

hadley:
I'm a little suspicious that something is going wrong with `guess_max` here

jennybc:
With respect to this change? Or in general and this PR made you read this code? I do have tests for guess_max btw.

hadley:
Why does changing something that should only affect 1000 rows have such a big impact on performance? Something is fishy.

jennybc:
Well, this sheet happens to have 100 *columns*.

But I can take a look.

hadley:
Ooooh, so it has to scan 100,000 cells. In that case, don't worry about it, but we might want to consider if there's a way to speed this up in the future.

jennybc:
Yeah I would generally like to profile the C++ code to see where the time goes.

PR #344, opened 2017-04-17
Simplify ifdef's around 64bit integers

jennybc:
Main point of this PR right now is to trigger ci builds


jennybc:
This works:

  * locally
  * Ubuntu Linux 16.04 LTS, R-release, GCC (r-hub)
  * Fedora Linux, R-devel, clang, fortran (r-hub)
  * Debian Linux, R-devel, GCC ASAN/UBSAN
  * all our existing travis checks (many flavours of linux & R and Mac OS)
  * win-builder

I don't have usable info from:

  * appveyor, because of a tidyverse traffic jam
  * Windows Server 2008 R2 SP1, R-oldrel, 32/64 bit (r-hub) because readxl always fails/hangs there and I don't know why

I'M GOING FOR IT.

jennybc:
Thanks @jimhester for this suggestion.

PR #384, opened 2017-09-01
Hide internal symbols on supported platforms

jeroen:
Rcpp now supports compiling with` $(C_VISIBILITY)` which will hide internal symbols from the dll. This reduces chances of symbol collision if other packages define the same symbols, for example because they also bundle rapidxml.


jennybc:
Thanks!

PR #398, opened 2017-10-19
Transform int to 0/1 before coercion to logical

### LINKED ISSUE #385: segfault when tabulating data incorrectly guessed as type "logical"
After reading the data in [this file](http://sfgov.org/citysurvey/sites/default/files/City%20Survey%20MASTER%20Data%201996-2017.xlsx), I get a segmentation fault whenever I try to aggregate most columns in the resulting `tbl_df`:

```
> library(readxl)
> df <- read_xlsx("City Survey MASTER Data 1996-2017.xlsx", sheet=2)
> table(df$parkvis)

 *** caught segfault ***
address 0x211ebd268, cause 'memory not mapped'

Traceback:
 1: unique.default(x, nmax = nmax)
 2: unique(x, nmax = nmax)
 3: factor(a, exclude = exclude)
 4: table(df$parkvis)

Possible actions:
1: abort (with core dump, if enabled)
2: normal R exit
3: exit R without saving workspace
4: exit R saving workspace
Selection: 1
R is aborting now ...
Segmentation fault: 11
```

The problem is probably related to the fact that, because there are so many missing values at the top of the dataset, this column (and a majority of the others in the dataset) were guessed as type "logical":

```
> class(df$parkvis)
[1] "logical"
> head(df$parkvis)
[1] NA NA NA NA NA NA
> tail(df$parkvis)
[1] TRUE TRUE TRUE TRUE TRUE TRUE
```

It turns out that if you specify the `col_types` correctly as "numeric", the data are perfectly readable without segfaulting:

```
> df <- read_xlsx("City Survey MASTER Data 1996-2017.xlsx", sheet=2, col_types="numeric")
> table(df$parkvis, useNA="always")

    1     2     3     4     5     6  <NA> 
 2027  2863  4931  6400  9325   828 11598 
```

```
> sessionInfo()
R Under development (unstable) (2017-09-07 r73219)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Sierra 10.12.6

Matrix products: default
BLAS: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRblas.0.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:

other attached packages:
[1] readxl_1.0.0

[1] compiler_3.5.0   tools_3.5.0      tibble_1.3.4     Rcpp_0.12.12    
[5] cellranger_1.1.0 rlang_0.1.2.9000
```

nacnudus:
Fixes #385

The clue to solving this issue was in [this twitter thread](https://twitter.com/wrathematics/status/921016665529356289).  Integers are coerced to logical in two places.  One place does

```cpp
SET_VECTOR_ELT(column, row, Rf_ScalarLogical(xcell->asInteger()));
```

where `Rf_ScalarLogical` transforms non-zero integers to 1 before coercing to logical.  But the other place does

```cpp
LOGICAL(column)[row] = xcell->asInteger();
````

Weirdly, the test passes even without this fix when run with `testthat::test()`.  To get the test to fail (in fact, segfault), checkout the previous commit, install, and from within the project directory do

```r
library(readxl)
library(testthat)
source("./tests/testthat/helper.R")
source("./tests/testthat/test-col-types.R")
```


jennybc:
Nice! Yes I had figured something like this was going on, but it's a big help to have it tracked down and fixed. Thanks!

I can reproduce almost everything you say, except this:

> Weirdly, the test passes even without this fix when run with `testthat::test()`.

I am running dev testthat and it does indeed catch it and correctly reports ` *** caught segfault ***`. So that's good news too.

jennybc:
@jimhester I'm going to merge this. It's super small but important and a good lesson. Will you speak up if this is somehow *not* the correct fix? The linked issue will illustrate the problem this fixes.

jennybc:
@nacnudus Will you make another PR with a bullet in NEWS?

src/XlsxCell.h
```diff
@@ -256,7 +256,7 @@ class XlsxCell {
     case CELL_NUMERIC:
     {
       rapidxml::xml_node<>* v = cell_->first_node("v");
-      return atoi(v->value());
+      return atoi(v->value()) != 0;
```
jimhester: This does not seem correct, won't this convert any numeric value to 0 or 1? You only want to do this for the `CELL_LOGICAL` case, not both, no?

Also I think it is maybe clearer to be more explicit and use

```c
return atoi(v->value()) != 0 ? 1 : 0;
```

But that is subjective.

jennybc: This is correct, in terms of how `asInteger` is actually used. It is only called when putting numeric data into logical cells.

But, yes, it suggests `asInteger` should perhaps be renamed to `asLogical`. Or I should have both. Or there should at least be a comment.

jimhester: I think it should be `asLogical` and I would use [`Rf_LogicalFromReal()`](https://github.com/wch/r-source/blob/d1f01dc790201e2e40c97e8da45c51c495874ffe/src/main/coerce.c#L94-L98) to do the conversion, which also handles the NaN case.

```c
return Rf_LogicalFromReal(atoi(v->value()), nullptr);
```

Issue #373, opened 2017-07-06
Error in read_fun(path = path, sheet = sheet, limits = limits, shim = shim,  :    std::bad_alloc


MidhunT:
I have an excel file with multiple sheets and each sheet more than 65000 rows. I get the above mentioned error with the latest version (1.0.0) of the package when I try `read_excel(path = x, sheet = y, col_names = T)` and when I try by downgrading the package with the same code it is ignoring one of the column. I have few timestamp columns so when I try by adding `col_types = c(rep("text"), 11)` it throwing me an error saying `Error in xls_cols(path, sheet, col_names = col_names, col_types = col_types,  : 
  col_names and col_types must have the same length` When I clearly have 11 columns. 
I'm using 16gb machine and i7 processor, I guess RAM size should not be an issue here.

jennybc:
So you're saying you have a `std::bad_alloc` error with v1.0.0 but NOT with the previous release? I.e. you can read the file w/o readxl giving you an explicit error with the previous release?

The problems you report re: ignoring one column and complaining about mismatched `col_names` and `col_types` are bugs fixed in v1.0.0.

Any chance you can provide this file to me? How was it created, i.e. was it Excel or some other tool?

MidhunT:
I'm afraid, I cannot share the file but yes the file was generated using Neware btsda software

jennybc:
There's little I can do w/o the file because there's not enough information to go on. Can you open this file in Excel and resave as `.xls` and as `.xlsx`? Does that change readxl's behaviour on it, i.e. can readxl import it after this resaving? That would be semi-informative.

I'm also still interested in confirmation or clarification on this:

> So you're saying you have a std::bad_alloc error with v1.0.0 but NOT with the previous release? I.e. you can read the file w/o readxl giving you an explicit error with the previous release?

MidhunT:
Yes
> So you're saying you have a std::bad_alloc error with v1.0.0 but NOT with the previous release? I.e. you can read the file w/o readxl giving you an explicit error with the previous release?

When I resave it as `.xlsx` file then it works but it did not work after resaving it as `.xls` and even if I save contents of only one column with 65536 rows (including column name) from one of the sheet of the existing file to new file with `.xls` extension it fails

jennybc:
[...]

PR #416, opened 2018-01-16
Avoid infinite loop in `XlsWorkSheet::loadCells`

vkapartzianis:
I had a quick look at issue #373 at the invitation of a colleague. The code definitely enters an infinite loop with the samples provided, so I've tried and avoid that.


jennybc:
Are you game to add a test and test sheet?

vkapartzianis:
The test file is ~6MB, is that OK?

jennybc:
Ooh no. Is that necessary? Sorry my mind is elsewhere but wanted to catch while your head was in right place, in case you can easily add test case.

vkapartzianis:
OK, I've tried to simplify the test case. It seems it boils down to just an XLS spreadsheet with 65536 rows.

This request doesn't resolve the real problem, in addition. I'll work on it some more.

[65536-rows.zip](https://github.com/tidyverse/readxl/files/1636985/65536-rows.zip)

jennybc:
> OK, I've tried to simplify the test case. It seems it boils down to just an XLS spreadsheet with 65536 rows.

Interesting. I've battled a few other cases of short vs long int in libxls. I mean, it's possible that I created this one. Or not.

vkapartzianis:
I've added another commit. It should handle all 65536-row cases and it includes a simple test.

Note that this should be a problem with XLS files only. XLSX sheets aren't limited to 65536 rows as far as I can tell.

jennybc:
Moving this to #432

Issue #278, opened 2017-03-01
Support reading from more general inputs


jennybc:
Right now readxl reads only from xlsx and xls files.

In the fullness of time, some functionality for supporting more general inputs will be pulled out of readr, at which point readxl can exploit that. We might implement an interim solution for some of these in the meantime.

This issue will cover all related feature requests (all of which I'm now closing):

  * #192 input via rawConnection
  * #183 URL support?
  * #115 align all read_* decompression and download behavior with read_delim()
  * #93 Read file directly from memory

jjcad:
Question: Along with this support for more general inputs, will that also signal a transition to `readr` style `col_types`? Having that kind of unity between the packages would be welcome.

jennybc:
The `col_types` transition is #198. I can't think of any reason that implementation of these two things would be inextricably tied together. They'd more likely happen around the same time because it meant I was working on a new release.

MichaelChirico:
[`fread`'s approach](https://github.com/Rdatatable/data.table/blob/master/R/fread.R#L44-L66) is to download the file & proceed to use plaid-old `fread` on that.

@jennybc If you're happy with that logic, I'd be happy to file a PR.

jennybc:
@MichaelChirico Yeah, that is an option we've contemplated re: "faking" read from a URL. I'm definitely interested in a PR that downloads to a temp file, then reads that.

jknowles:
Does this include reading files from within a zip file using the `unz()` command like below:

```
zzz <- read_xlsx(unz("data/myzip.zip", "one_of_many.xlsx"))
```
This would be a handy option for allowing users to access compressed files.

jennybc:
@jknowles I rather doubt it. It's especially hard to see that working for xlsx, where we explicitly unpack *that* into individual XML files.

jennybc:
I started to tackle the URL piece of this as part of #454 and there is also another separate PR for URL download in #426. But Slack discussion has convinced me I really should tackle all of these together and soon-ish. But not for this week's release, which gets the security-patched libxls out there.

Notes from team discussion:

  * Even though `download.file()` is better behaved now wrt https, it may still be wise to prefer `curl::curl_download()`.
[...]

PR #426, opened 2018-02-20
read_excel reads from URL, part of #278

MichaelChirico:
Basic idea's done. My approach is to augment `check_file` to more flexibly detect a URL if `!file.exists(path)`.

Use of `assign` to overwrite `path` is a bit wonky but I couldn't see how else to do it without rewriting `check_format` to return both the updated `path` and the inferred format. I considered re-factoring everything but a quick/logical way of doing so didn't come to me right away.

I am not clear why `read_excel_` goes through `check_file` again, given that it's a hidden function and `check_file` should always be run by the time `read_excel_` is. If we drop the `check_format` from `read_excel_`, we could skip the `assign('path')` kludge since it only exists to avoid downloading the file again at this juncture.

As far as testing, I don't know what's the proper way to test reading from a URL, so I didn't go too crazy and only added one test. It's self-referential, which feels... risky, but I weighed that against the option of picking some Excel file "in the wild" and hoping the file remains there in perpetuity; current implementation seems better than that, but I'm open to ideas.

Also open to using an `option` to change the `verbose` behavior of the file download, or setting the default to `FALSE`.


MichaelChirico:
Just down to a Codecov failure. Can add tests as needed, just not sure the right way to do so for this case.

jennybc:
Thanks for this! I am refactoring the pre-processing of `path` more generally, so just a heads up that this may get swept up into the larger effort (#454). Regardless, I'll certainly acknowledge your effort to things moving on this front.

MichaelChirico:
@jennybc cool. happy to help hammer this into the redesign if you think it makes sense to delegate; just let me know.

jennybc:
Thanks again. I've decided to hold off on handling URLs for this release. Notes on my thinking are [here](https://github.com/tidyverse/readxl/issues/278#issuecomment-381720113).

PR #429, opened 2018-03-07
Handle xlsx files with missing v node.

rstub:
We have encountered xlsx files where some cells look like `<c r="H9" s="12" t="str"><f>F9*G9</f></c>` leading to a crash when reading the file. Unfortunately I cannot share these sheets but I was able to reproduce the issue by manually editing (see new test sheet). The proposed patch is one way to ensure that `v` is not dereferenced when it is `NULL`.


jennybc:
I cannot open the test xlsx file you include in this PR with Excel itself, which is basically a requirement for any special things we do re: xlsx produced by a 3rd party software.

screen shot 2018-03-16 at 4 40 22 pm

If Excel can't read it, it's not clear that readxl needs to read it. I'm afraid we're stuck until we can have an example sheet that Excel can open and that has cells with a missing v node.

rstub:
Interesting. I do not have Excel directly available, but I will see if I can find someone who can check the original file for me. If that works in Excel without repairing the file first, then I might be able to provide a better sample file. Both LibreOffice and Google Docs have no problem opening the test file.

However, the screen shot from `readxl` upon opening the test file is like this:

bildschirmfoto vom 2018-03-19 11-42-15

But only in the development version of `readxl`, since the problematic behavior was introduced in #404. So even if you do not like test files that do not work in Excel, I still believe that checking for `NULL` is mandatory.

rstub:
With the help of https://github.com/vindvaki/xlsx-validator I was able to alter the test file in such a way that it works in Excel but crashes R when read with the current development version of `readxl`.

jennybc:
So what do you expect the contents of this cell to be once your read it into R?

It's a very peculiar situation to have a formula cell and to NOT have the computed result. I guess this 3rd party tool writes such cells and ... assumes that result will be computed the first time it's opened and resaved in Excel?

I suppose it has to come in as NA, i.e. we just proceed as if the cell didn't appear in the file. I can't see what else one would do.

jennybc:
OK I will push a few tweaks/additions here tomorrow and then merge. Thanks!

PR #436, opened 2018-03-23
Change na parameter description to match readr

kwstat:
readr/R/tokenizer.R says:

#' @param na Character vector of strings to interpret as missing values. Set this
#'   option to `character()` to indicate no missing values.


jennybc:
Thanks!

PR #437, opened 2018-03-24
Parse _rels/.rels to get paths to package parts

### LINKED ISSUE #435: Be able to read sheets produced by the Los Angeles Police Department
https://source.opennews.org/articles/how-we-found-new-patterns-la-homeless-arrest/

I have a specimen now, thanks to @underthecurve.

The xlsx produced by LAPD is a very literal implementation of the example on pp65-66 of 5th edition of ECMA:

> Example: The following package represents the minimal conformant SpreadsheetML package as defined by ECMA-376

jennybc:
Fixes #435 Be able to read sheets produced by the Los Angeles Police Department

The xlsx produced by LAPD is a very literal implementation of the example on pp65-66 of 5th edition of ECMA:

> Example: The following package represents the minimal conformant SpreadsheetML package as defined by ECMA-376

I used this as motivation to make a major upgrade to how readxl parses the package structure. Basically, readxl previously did not parse it but, rather, assumed a very common default. And the LAPD sheets show that other structures are possible (Excel can open the LAPD files, btw). Now we can read LAPD sheets, while still being able to read all the other weird 3rd party sheets in our test collection 🤞

The main feature of the LAPD sheets is that they do not use an `xl/` subdirectory, whereas every other xlsx we've seen before does. Browsable examples:

  * LAPD sheets do not use an `xl/` subdirectory: [excelgesis/los-angeles-arrests-xlsx](https://jennybc.github.io/excelgesis/los-angeles-arrests-xlsx/index.html)
  * [excelgesis/embedded-chartsheet](https://jennybc.github.io/excelgesis/embedded-chartsheet/index.html) is a good example of a "normal" xlsx with multiple worksheets and a chartsheet. It has an `xl/` subdirectory.
  * [excelgesis/Ekaterinburg_IP_9](https://jennybc.github.io/excelgesis/Ekaterinburg_IP_9/index.html) is a valuable example of yet another type of 3rd party xlsx. It uses `xl/` but insists on giving paths in `/xl/foo/bar` form in all the relationship files, which is peculiar (very different from embedded-chartsheet example). It's also completely in Cyrillic, which is extra fun.

### DIFF
--- a/src/XlsxWorkBook.h
+++ b/src/XlsxWorkBook.h
@@ -10,16 +10,58 @@
 
 class XlsxWorkBook {
 
-  // holds objects related to sheet position, name, Id, and xml target file
-  class SheetRelations {
+  // holds objects related to package parts, such file paths, but also
+  // worksheet position, name, and id
+  class PackageRelations {
+    // non-worksheet package parts
+    std::map<std::string, std::string> part_;
+
+    // worksheets
     int n_;
     Rcpp::CharacterVector names_;
     Rcpp::CharacterVector id_;
     std::map<std::string, std::string> target_;
 
+    // populate workbook element of part_
+    void parse_package_rels(const std::string& path) {
+      std::string rels_xml_file = zip_buffer(path, "_rels/.rels");
+      rapidxml::xml_document<> rels_xml;
+      rels_xml.parse<rapidxml::parse_strip_xml_namespaces>(&rels_xml_file[0]);
+
+      rapidxml::xml_node<>* relationships = rels_xml.first_node("Relationships");
+      if (relationships == NULL) {
+        Rcpp::stop("Spreadsheet has no package-level relationships");
+      }
+
+      std::map<std::string, std::string> scratch;
+      for (rapidxml::xml_node<>* relationship = relationships->first_node();
+           relationship; relationship = relationship->next_sibling()) {
+
+        rapidxml::xml_attribute<>* type = relationship->first_attribute("Type");
+        rapidxml::xml_attribute<>* target = relationship->first_attribute("Target");
+
+        if (type != NULL && target != NULL) {
+          // keys are derived by abbreviating Type
+          // in XML: http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument
+          // key is just the last part: officeDocument
+          scratch[baseName(type->value())] = target->value();
+        }
+      }
+
+      // identify and store the workbook part = the main part
+      // ECMA-376 Part 1 section 8.5 SpreadsheetML
+      std::map<std::string, std::string>::iterator m = scratch.find("officeDocument");
+      if (m == scratch.end()) {
+        Rcpp::stop("No workbook part found");
+      }
+      // store 'xl/workbook.xml', not '/xl/workbook.xml' (rare, but have seen)
+      part_["officeDocument"] = deSlash(m->second);
+    }
+
     // populates n_, names_, id_
     void parse_workbook(const std::string& path) {
-      std::string workbookXml = zip_buffer(path, "xl/workbook.xml");
+      std::string workbookXml = zip_buffer(path, part_["officeDocument"]);
+
       rapidxml::xml_document<> workbook;
       workbook.parse<rapidxml::parse_strip_xml_namespaces>(&workbookXml[0]);
 
@@ -59,7 +101,13 @@ class XlsxWorkBook {
 
     // populates target_
     void parse_workbook_rels(const std::string& path) {
-      std::string rels_xml_file = zip_buffer(path, "xl/_rels/workbook.xml.rels");
+      const std::string workbook_path = part_["officeDocument"];
+      const std::string workbook_dir = dirName(workbook_path);
+      std::string rels_xml_path =
+        workbook_dir + "/_rels/" + baseName(workbook_path) + ".rels";
+      rels_xml_path = deSlash(rels_xml_path);
+      std::string rels_xml_file = zip_buffer(path, rels_xml_path);
+
       rapidxml::xml_document<> rels_xml;
       rels_xml.parse<rapidxml::parse_strip_xml_namespaces>(&rels_xml_file[0]);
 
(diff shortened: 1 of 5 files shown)


src/XlsxWorkBook.h
```diff
@@ -10,16 +10,58 @@
 
 class XlsxWorkBook {
 
-  // holds objects related to sheet position, name, Id, and xml target file
-  class SheetRelations {
+  // holds objects related to package parts, such file paths, but also
+  // worksheet position, name, and id
+  class PackageRelations {
```
jennybc: New name reflects that we now store paths for non-(work)sheet package parts.

src/XlsxWorkBook.h
```diff
@@ -10,16 +10,58 @@
...
+
+    // worksheets
     int n_;
     Rcpp::CharacterVector names_;
     Rcpp::CharacterVector id_;
     std::map<std::string, std::string> target_;
 
+    // populate workbook element of part_
+    void parse_package_rels(const std::string& path) {
+      std::string rels_xml_file = zip_buffer(path, "_rels/.rels");
```
jennybc: readxl previously ignored this file and basically hard-wired the most prevalent structure for a SpreadsheetML package. The LAPD sheets demonstrate this is unsafe and it is more correct to parse this file.

src/XlsxWorkBook.h
```diff
@@ -10,16 +10,58 @@
...
+      rapidxml::xml_document<> rels_xml;
+      rels_xml.parse<rapidxml::parse_strip_xml_namespaces>(&rels_xml_file[0]);
+
+      rapidxml::xml_node<>* relationships = rels_xml.first_node("Relationships");
+      if (relationships == NULL) {
+        Rcpp::stop("Spreadsheet has no package-level relationships");
+      }
+
+      std::map<std::string, std::string> scratch;
```
jennybc: I'm parsing the whole thing because it's easy and one day we might care about the other files covered here. But I'm parsing into a `scratch` map, because for now I only care about worksheet part. That is identified and stored properly in `part_` right after this loop.

src/XlsxWorkBook.h
```diff
@@ -59,7 +101,13 @@ class XlsxWorkBook {
 
     // populates target_
     void parse_workbook_rels(const std::string& path) {
-      std::string rels_xml_file = zip_buffer(path, "xl/_rels/workbook.xml.rels");
+      const std::string workbook_path = part_["officeDocument"];
+      const std::string workbook_dir = dirName(workbook_path);
+      std::string rels_xml_path =
+        workbook_dir + "/_rels/" + baseName(workbook_path) + ".rels";
+      rels_xml_path = deSlash(rels_xml_path);
```
jennybc: LAPD sheets teach us that an `xl/` subdirectory is not necessarily used, i.e. `workbook_dir` can be `""`. So we must construct this path.

src/XlsxWorkBook.h
```diff
@@ -109,15 +174,25 @@ class XlsxWorkBook {
...
       return it->second;
     }
-  }; // end of class SheetRelations
+
+    std::string part(std::string type) const {
+      std::map<std::string, std::string>::const_iterator it = part_.find(type);
+      if (it == part_.end()) {
+        // e.g., sharedStrings may not be present
+        // downstream functions should handle non-existent path
+        return "";
```
jennybc: Is there a better sentinel value? I don't want to `stop()` if the key is not found.

src/utils.h
```diff
@@ -145,4 +145,25 @@ inline std::string trim(const std::string& s) {
...
+    return path;
+  }
+  return path.substr(found + 1);
+}
+
+inline std::string deSlash(const std::string& s) {
+  size_t start = s.find_first_not_of("/");
+  return s.substr(start);
+}
+
```
jennybc: Oh to have access to path handling functions. Do these look ok?

src/XlsxWorkBook.h
```diff
@@ -10,16 +10,58 @@
 
 class XlsxWorkBook {
 
-  // holds objects related to sheet position, name, Id, and xml target file
-  class SheetRelations {
+  // holds objects related to package parts, such file paths, but also
+  // worksheet position, name, and id
+  class PackageRelations {
+    // non-worksheet package parts
+    std::map<std::string, std::string> part_;
```
jennybc: A new private map to hold paths of non-worksheet package parts.

src/utils.h
jimhester: I think calling `deSlash` something like `removeLeadingSlash` would make it more clear what it does, I was not sure if `deSlash` was supposed to remove the leading slashes, trailing slashes or all slashes.

Also it does not handle the case where the input consists of only slashes, e.g. what happens if `find_first_not_of` returns `std::string::npos`.

src/utils.h
```diff
@@ -145,4 +145,25 @@ inline std::string trim(const std::string& s) {
...
+  }
+  return path.substr(0, found);
+}
+
+inline std::string baseName(const std::string& path) {
+  std::size_t found = path.find_last_of("/");
+  if (found == std::string::npos) {
+    return path;
+  }
+  return path.substr(found + 1);
```
jimhester: What about it `path` == `/`?

src/XlsxWorkBook.h
jimhester: I think the empty string is fine, you could theoretically use something like `\0`, but I think it is more trouble than it is worth.

src/utils.h
jennybc: This seems to be OK because `found + 1` is equal to the length of `path` and we get the empty string back.

From [std::string::substr](http://www.cplusplus.com/reference/string/string/substr/):

> *pos* If this is equal to the string length, the function returns an empty string.

I also doublechecked in a little example.

src/utils.h
jennybc: OK yes. I've now addressed the case of input that consists entirely of slashes. Also renamed to `removeLeadingSlashes()` for complete truth in advertising.

PR #452, opened 2018-04-15
Switch to tibble::set_tidy_names() from tibble::repair_names()

### LINKED ISSUE #357: column names have double underscores in them when col_names = F after latest update
After #98 was released I am getting "__" in the names of the columns when col_names = F.  https://github.com/tidyverse/readxl/releases it says:
Now, the default column names are X__1, X__2, X__3. I don't understand why there is a double underscore. #98 only discusses fixing the numbers not adding underscores.

Thank you

`> sessionInfo()
R version 3.3.3 (2017-03-06)

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:

other attached packages:
[1] readxl_1.0.0

[1] tools_3.3.3      tibble_1.3.0     Rcpp_0.12.10     cellranger_1.1.0`

jennybc:
Resolves #357

Recall previous discussion about name repair, that lead to `tibble::set_tidy_names()`: https://github.com/tidyverse/tibble/issues/217

I would really like to make this change. And I predict that revdep checks will reveal few/no problems with this move in *packages*.

But I also know from the previous release that any change here generates a lot of fretting and issues, as people have scripts that are hardwired to whatever the previous name repair strategy was. It also always brings up the usual suggestions to clean up nonsyntactic names. Quoting @hadley from thread linked above:

> I'm not sure what we should do with respect to backward compatibility. I suspect few packages depend on this behaviour; it's going to be more user code. I think as long as there's a clear way to get to the old behaviour it shouldn't cause much hassle (especially since we're now describing what's happening)

Thoughts on whether and how to proceed?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # readxl 1.0.0.9000
 
+* Missing or duplicated column names are now repaired with `tibble::set_tidy_names()` in `read_excel()` and friends. `set_tidy_names()` is intended to implement a standard across the tidyverse packages for repairing missing and duplicated names. Its design is discussed in [tidyverse/tibble#217](https://github.com/tidyverse/tibble/issues/217). (PR here, #357)
+
 * Embedded libxls has been updated to address security vulnerabilitities identified in late 2017 (#441, #442).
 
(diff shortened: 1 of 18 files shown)


jennybc:
Will reopen with nicer history

PR #453, opened 2018-04-15
Switch to tibble::set_tidy_names() from tibble::repair_names()

### LINKED ISSUE #357: column names have double underscores in them when col_names = F after latest update

jennybc:
Resolves #357

Recall previous discussion about name repair, that lead to `tibble::set_tidy_names()`: https://github.com/tidyverse/tibble/issues/217

I would really like to make this change. And I predict that revdep checks will reveal few/no problems with this move in *packages*.

But I also know from the previous release that any change here generates a lot of upset and issues, as people have scripts that are hardwired to whatever the previous name repair strategy was. It also always brings up the usual suggestions to clean up nonsyntactic names. Quoting @hadley from thread linked above:

> I'm not sure what we should do with respect to backward compatibility. I suspect few packages depend on this behaviour; it's going to be more user code. I think as long as there's a clear way to get to the old behaviour it shouldn't cause much hassle (especially since we're now describing what's happening)

Thoughts on whether and how to proceed?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # readxl 1.1.0.9000
 
+* Missing or duplicated column names are now repaired with `tibble::set_tidy_names()` in `read_excel()` and friends. `set_tidy_names()` is intended to encourage name repair that is more principled and consistent, across multiple tidyverse packages. Its design is discussed in [tidyverse/tibble#217](https://github.com/tidyverse/tibble/issues/217). (#357, #453)
+
+  - Example of change a user will see: consider a spreadsheet with three columns, one unnamed and two named `x`.
+  - Column names in readxl > 1.1.0: `..1`, `x..2`, `x..3`
+  - Column names in readxl <= 1.1.0: `X__1`, `x`, `x__1`
+
 # readxl 1.1.0
 
 * `read_excel()` and `excel_sheets()` associate a larger set of file extensions with xlsx and are better able to guess the format of a file with a nonstandard or missing extension. This is about deciding whether to treat a file as xls or xlsx. (#342, #411, #457)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -27,7 +27,7 @@ LinkingTo: Rcpp
 Imports:
     cellranger,
     Rcpp (>= 0.12.12),
-    tibble (>= 1.1)
+    tibble (>= 1.3.1)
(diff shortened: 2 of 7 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # readxl 1.0.0.9000
 
+* Missing or duplicated column names are now repaired with `tibble::set_tidy_names()` in `read_excel()` and friends. `set_tidy_names()` is intended to implement a standard across the tidyverse packages for repairing missing and duplicated names. Its design is discussed in [tidyverse/tibble#217](https://github.com/tidyverse/tibble/issues/217). (PR here, #357)
```
hadley: You might also want to include as a heading under a "Breaking changes" section where you describe the issue from the symptoms (i.e. how have names changed), and offer a work around in code.

jennybc: Roger re: "Breaking changes"

> offer a work around in code

Well that's the problem. I don't really offer one right now. To offer a nice one ... I think I'd need to add an argument that affects name repair. Is it worth it?

The only workaround possible with the current strategy would be a *post hoc* regex-based renaming strategy, given what I know about the old and new repaired names.

jennybc:
I'm going to hold off on this but will foreshadow in NEWS/blog post.

Will do shortly after release, so people have time to adjust. Would be obvious to combine in a CRAN release with col spec improvements.

krlmlr:
[*pkgconfig*](https://github.com/r-lib/pkgconfig) works for me in these situations: We could start with opt-in even before CRAN release, and just switch the default after CRAN release.

PR #454, opened 2018-04-16
Refactor path pre processing

jennybc:
WIP


R/read_excel.R
```diff
@@ -81,10 +82,16 @@ NULL
 #'
 #' # Get a preview of column names
 #' names(read_excel(readxl_example("datasets.xlsx"), n_max = 0))
+#'
+#' \dontrun{
+#' # Read from a URL
+#' read_excel("https://github.com/tidyverse/readxl/blob/master/inst/extdata/datasets.xlsx?raw=true", n_max = 3)
```
MichaelChirico: Oh never knew about the `?raw=true` tag. convenient 👍

R/utils.R
```diff
@@ -1,6 +1,16 @@
 check_file <- function(path) {
+  if (!is_string(path)) {
+    stop("`path` must be a string", call. = FALSE)
+  }
+
+  if (grepl("^((http|ftp)s?|sftp)://", path)) {
```
MichaelChirico: `fread` built in support for `file://` URLs; this is helpful for testing, since such URLs can be treated exactly like web URLs from unit testing perspective (i.e., `download.file` can be dispatched) but don't require internet connection

jennybc:
I will shortly take another crack at this, via close and re-open or force push. It will decouple (remove, actually) URL handling from expanding the universe of xlsx-ish files we will attempt to read.

PR #459, opened 2018-04-18
Use a pragma to silence the C99 extensions warnings

jimhester:
This also reverts changing the flexible array members to single element
arrays.

### DIFF
--- a/src/ColSpec.h
+++ b/src/ColSpec.h
@@ -1,6 +1,9 @@
 #ifndef READXL_COLSPEC_
 #define READXL_COLSPEC_
 
+#pragma GCC diagnostic ignored "-Wc99-extensions"
+#pragma clang diagnostic ignored "-Wc99-extensions"
+
 #include <Rcpp.h>
 #include <libxls/xls.h>
 #include "StringSet.h"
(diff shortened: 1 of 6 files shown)


jennybc:
THANK YOU.

We might still have a NOTE? I see this in the [travis log](https://travis-ci.org/tidyverse/readxl/jobs/368299213#L1135) for r-devel.

```
Files which contain pragma(s) suppressing diagnostics:
  ‘src/ColSpec.h’ ‘src/XlsWorkBook.cpp’
```

Does that surprise you?

Seems like we have two choices:

  * Proceed with this ☝️ note
  * Leave the zero-length arrays and flexible array members (and I undo my existing remediations) and have that note

Regardless, we explain the situation in CRAN comments, as discussed elsewhere.

jennybc:
I'm going to ship current master and this off to win-builder for a head-to-head comparison.

jimhester:
Ah I forgot that NOTE was only on R-devel, I should be able to remove it as well one sec.

jimhester:
Should be fixed by https://github.com/tidyverse/readxl/pull/459/commits/de0f79b7f98b7faa0842d1a08f9192f1a0baef14

jennybc:
Nice! I think this settles the debate between this PR and #460, in favour of this. I will still do a full round of win builder before I submit.

PR #460, opened 2018-04-18
Backout array hacks

jennybc:
This is a version that removes all readxl workarounds re: zero size arrays and flexible array members.

It is looking increasingly likely that we must submit with some sort of NOTE and we are now choosing between compiler warnings like these:

```
./libxls/xlsstruct.h:126:10: warning: flexible array members are a C99 feature [-Wc99-extensions]
    char        name[];
                ^

./libxls/xlsstruct.h:243:18: warning: zero size arrays are an extension [-Wzero-length-array]
    BYTE        strings[0];
                        ^
```

or this NOTE:

```
Files which contain pragma(s) suppressing diagnostics:
  ‘src/ColSpec.h’ ‘src/XlsWorkBook.cpp’
```

associated with the solution in #459.

### DIFF
--- a/src/libxls/xlsstruct.h
+++ b/src/libxls/xlsstruct.h
@@ -194,7 +194,7 @@ typedef struct MULRK
 	struct {
 		WORD	xf;
 		DWORD_UA value;
-	}		rk[1]; // readxl
+	}		rk[];
 	//WORD	last_col;
 }
 MULRK;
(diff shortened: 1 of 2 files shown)


jennybc:
It's clear we are on a different and better path now, within readxl and in collaboration with upstream libxls development.

PR #465, opened 2018-04-24
Don't push empty string onto shared string table

### LINKED ISSUE #463: xls regression: can't read some xls with v1.1.0 that worked with v1.0.0
Please briefly describe your problem and what output you expect. If you have a question, please don't use this form. Instead, ask on <https://stackoverflow.com/> or <https://community.rstudio.com/>.

Please include a minimal reproducible example (AKA a reprex). If you've never heard of a [reprex](http://reprex.tidyverse.org/) before, start by reading <https://www.tidyverse.org/help/#reprex>.

---

Brief description of the problem
I used to use readxl to read xls and xlsx files, however this time when I run my code 
sheetlist <- readxl::excel_sheets("EIU_Data.xls")
Error in xls_sheets(path) : Failed to open EIU_Data.xls

however when I used a different version in xlsx
temp <- readxl::read_excel("EIU_Data.xlsx", sheet = 1, col_names = FALSE, na = "n.a.")
na becomes AF
Showning Data is no in the file, so Data is inconsistent.

Please advise.

```r
# insert reprex here
```
[EIU.zip](https://github.com/tidyverse/readxl/files/1936625/EIU.zip)

### LINKED ISSUE #467: Shared string table parsing in some xlsx from Economist Intelligence Unit
I want to separate #463 into an xlsx problem (this issue) and an xls problem (how I will use #463).

PR and #463 give details.

jennybc:
Fixes #467

### DIFF
--- a/src/XlsxWorkBook.h
+++ b/src/XlsxWorkBook.h
@@ -260,8 +260,9 @@ private:
     for (rapidxml::xml_node<>* string = sst->first_node();
          string; string = string->next_sibling()) {
       std::string out;
-      parseString(string, &out);    // missing strings are treated as empty
-      stringTable_.push_back(out);
+      if (parseString(string, &out)) { // missing strings are treated as empty
+        stringTable_.push_back(out);
+      }
     }
   }
 
--- a/src/XlsxWorkBook.h
+++ b/src/XlsxWorkBook.h
@@ -260,7 +260,7 @@ private:
     for (rapidxml::xml_node<>* string = sst->first_node();
          string; string = string->next_sibling()) {
       std::string out;
-      if (parseString(string, &out)) { // missing strings are treated as empty
+      if (parseString(string, &out)) {
         stringTable_.push_back(out);
       }
     }


jennybc:
@jimhester Do you see any logical flaw in my fix? I'm not sure why we ever pushed `out` onto the string table if we were unable to find any of the nodes documented to harbour actual strings. I see no downside to the proposed change and all tests pass everywhere.

I did some print debugging on two versions of the xlsx provided in #463 and, for some reason that still eludes me, the OP's xlsx **does** cause a visit to a phantom node and was causing the first entry of the string table to be the empty string. The same xlsx file, when opened and resaved in Excel, does not have such a visit and does not contaminate the string table. Evidence = the first line of `x <- ...` (no "t node found") versus `y <- ...` below.

```
> x <- read_excel("~/Downloads/EIU/EIU_Data.xlsx", n_max = 1)
in parseString, exiting
in parseString, t node found, exiting
in parseString, t node found, exiting
in parseString, t node found, exiting
in parseString, t node found, exiting
> y <- read_excel("~/Downloads/EIU/EIU_Data-RESAVED.xlsx", n_max = 2)
in parseString, t node found, exiting
in parseString, t node found, exiting
in parseString, t node found, exiting
in parseString, t node found, exiting
in parseString, t node found, exiting
```

Here is the top of `sharedStrings.xml` in each case. Somehow this seemingly meaningless difference matters.

---
`EIU_Data/xl/sharedStrings.xml`

```
<?xml version="1.0" encoding="utf-8" standalone="yes"?><sst xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" count="253104" uniqueCount="3377">﻿<si><t>DGDP</t></si><si><t>No of countries : 202</t>
```

---
`EIU_Data-RESAVED/xl/sharedStrings.xml`

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<sst xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" count="253104" uniqueCount="3377"><si><t>DGDP</t></si><si><t>No of countries : 202</t></si>
```

src/XlsxWorkBook.h
```diff
@@ -260,8 +260,9 @@ class XlsxWorkBook {
     for (rapidxml::xml_node<>* string = sst->first_node();
          string; string = string->next_sibling()) {
       std::string out;
-      parseString(string, &out);    // missing strings are treated as empty
-      stringTable_.push_back(out);
+      if (parseString(string, &out)) { // missing strings are treated as empty
```
jennybc: I think I'm going to delete or reword this confusing comment too

PR #477, opened 2018-05-05
Make sure path is in native encoding

### LINKED ISSUE #370: read_excel on Windows fails to open a file from a UTF-8 encoded path containing special characters
Using readxl 1.0.0 in R 3.4.0 for Windows, I recently came across an unexpected error when trying to open an Excel file containing a non-ASCII character in the file name. Depending on the encoding of the path variable, the import either works or fails:

```R
> library(readxl)

> # Copy an example file to a path including a special character
> my.path <- "Tür.xlsx"

> file.copy(system.file("extdata", "datasets.xlsx", package="readxl"), my.path, overwrite = TRUE)
[1] TRUE

> # Try to open file with path in native encoding
> read_excel(enc2native(my.path))
# A tibble: 150 x 5
   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
          <dbl>       <dbl>        <dbl>       <dbl>   <chr>
 1          5.1         3.5          1.4         0.2  setosa
 2          4.9         3.0          1.4         0.2  setosa
 3          4.7         3.2          1.3         0.2  setosa
 4          4.6         3.1          1.5         0.2  setosa
 5          5.0         3.6          1.4         0.2  setosa
 6          5.4         3.9          1.7         0.4  setosa
 7          4.6         3.4          1.4         0.3  setosa
 8          5.0         3.4          1.5         0.2  setosa
 9          4.4         2.9          1.4         0.2  setosa
10          4.9         3.1          1.5         0.1  setosa
# ... with 140 more rows

> # Try to open file with path in UTF-8 encoding
> read_excel(enc2utf8(my.path))
Error in read_fun(path = path, sheet = sheet, limits = limits, shim = shim,  : 
  Evaluation error: zip file 'TÃ¼r.xlsx' cannot be opened.
```

### LINKED ISSUE #476: CP1251 char set in file name
On Windows 7 and R version 3.5.0 (3.4.4 works without errors)
with the file included in ZIP (I can not upload .XLS files in this form).
[ISSUE.zip](https://github.com/tidyverse/readxl/files/1970104/ISSUE.zip)
`library(readxl)`
`fName <- "ТЕСТ.xls"`
`RES <- read_excel(fName, sheet=1, col_names=TRUE, col_types=NULL, na=" ", skip=0)`
Error message is:
`Error in read_fun(path = path, sheet_i = sheet, limits = limits, ...)`
`Filed to open ТЕСТ.xls`

!?!File with .XLSX extension works without errors:
`fName <- "ТЕСТ.xlsx"`

But I always prefer Excel 2003 .XLS format.

Session Info:
R version 3.5.0 (2018-04-23)

Matrix products: default

locale:
[1] LC_COLLATE=Russian_Russia.1251  LC_CTYPE=Russian_Russia.1251   
[3] LC_MONETARY=Russian_Russia.1251 LC_NUMERIC=C                   
[5] LC_TIME=Russian_Russia.1251    

attached base packages:

[1] compiler_3.5.0

jennybc:
Fixes #476 CP1251 char set in file name

Fixes #370 read_excel on Windows fails to open a file from a UTF-8 encoded path containing special characters

This is a new problem seen with R 3.5, we believe as a result of measures taken in response to [Bug 17120 ](https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=17120):

In a non-UTF-8 locale (basically Windows?), `path.expand()` returns a UTF-8 encoded path, even for non UTF-8 input. readxl pre-processes xls paths via `normalizePath()`, which calls `path.expand()`. And then libxls uses `fopen()`, which expects the path in the native encoding.

We have two choices:

  1. Stop using `normalizePath()` on xls files. It's not clear to me why we do this. This code goes back to Day 3 of readxl development, in 2015 fc076c463dca90407e03a8c988df6c65936051bb. The history doesn't indicate if there is a reason to use `normalizePath()`. I explored "stop calling `normalizePath()`" in 945178d41dbef4b02a0ce56c7fffc1793d2eaed6, which passes tests. If we choose this, I will obviously remove more even code vs. comment it out.
  2. Keep using `normalizePath()` on xls files, but process immediately with `enc2native()`.  Explored in 8a8c06136c9254b029368c24dc34915f49af6d0c, which also passes tests.

Either way, I think we should consider incorporating the new test. However I decided `skip_on_cran()` might be wise.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,19 @@
 # readxl 1.1.0.9000
 
+## Breaking change
+
 * Missing or duplicated column names are now repaired with `tibble::set_tidy_names()` in `read_excel()` and friends. `set_tidy_names()` is intended to encourage name repair that is more principled and consistent, across multiple tidyverse packages. Its design is discussed in [tidyverse/tibble#217](https://github.com/tidyverse/tibble/issues/217). (#357, #453)
 
   - Example of change a user will see: consider a spreadsheet with three columns, one unnamed and two named `x`.
   - Column names in readxl > 1.1.0: `..1`, `x..2`, `x..3`
   - Column names in readxl <= 1.1.0: `X__1`, `x`, `x__1`
+  
+## Other changes
+
+* Path handling (#477):
+
+  - `.xls` paths are no longer normalized. (#476 xls)
+  - All paths are explicitly converted to the native encoding via `enc2native()`  (#370)
 
 # readxl 1.1.0
 
--- a/R/read_excel.R
+++ b/R/read_excel.R
@@ -152,7 +152,7 @@ read_excel_ <- function(path, sheet = NULL, range = NULL,
   tibble::set_tidy_names(
     tibble::as_tibble(
       read_fun(
-        path = path, sheet_i = sheet,
+        path = enc2native(path), sheet_i = sheet,
         limits = limits, shim = shim,
         col_names = col_names, col_types = col_types,
         na = na, trim_ws = trim_ws, guess_max = guess_max
--- a/src/XlsWorkBook.h
+++ b/src/XlsWorkBook.h
@@ -6,13 +6,6 @@
 #include "ColSpec.h"
 #include "utils.h"
 
-inline std::string normalizePath(std::string path) {
-  Rcpp::Environment baseEnv = Rcpp::Environment::base_env();
-  Rcpp::Function normalizePath = baseEnv["normalizePath"];
-  Rcpp::Function enc2native = baseEnv["enc2native"];
-  return Rcpp::as<std::string>(enc2native(normalizePath(path, "/", true)));
-}
-
 class XlsWorkBook {
 
   // common to Xls[x]WorkBook
@@ -27,7 +20,7 @@ class XlsWorkBook {
 public:
 
   XlsWorkBook(const std::string& path) {
-    path_ = normalizePath(path);
+    path_ = path;
 
     xls::xlsWorkBook* pWB_ = xls::xls_open(path_.c_str(), "UTF-8");
     if (pWB_ == NULL) {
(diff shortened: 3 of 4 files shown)


yutannihilation:
FYI: using `Rf_translateChar()` may be another choice. (e.g. https://github.com/tidyverse/readr/pull/838/files)

IMHO, converting `Rcpp::String` directly to `std::string` might not do the right thing, as it doesn't care about the encoding, whereas base R provides useful functions that considers the encoding (`Rf_translateChar()`, `Rf_translateChar0()` and `Rf_translateCharUTF8()`). But, I don't know how to do this with Rcpp properly... (I've asked on [RStudio Community](https://community.rstudio.com/t/any-resources-to-learn-about-how-to-work-with-character-encodings-with-rcpp/7938), but no one has answered so far)

jimhester:
There is no `Rcpp::String` created in this case, the results from the R function calls are just `SEXP` objects and coercing to std::string should just do the equivalent of `std::string(CHAR())`.

yutannihilation:
> There is no `Rcpp::String` created in this case

Thanks for pointing out, I was wrong at this point. I meant, it might be also good if we can use the encoding attribute in C++ codes. But, as it seems difficult, just ignore me. Sorry for the noise...

jennybc:
@hadley Do you have any memory of why xls paths are sent through `normalizePath()`? But xlsx are not.

@jimhester Which approach do you think I should take?

I now introduce a 3rd option:

We should normalize all paths and do `enc2native()` on the R side.

Why?
I now see there was a pre-existing issue about **xlsx** and encoding of the path: #370
Also a new xls one that's probably this: #478

jimhester:
I think doing `enc2native(normalizePath())` on the R side for xls and xlsx seems the best option.

hadley:
I have no recollection of why there is a different.

jennybc:
I went with a different option:

  * Add a test with a filename containing non-ASCII characters. Apply it to both xls and xlsx.
  * Stop normalizing xls paths.
  * Apply `enc2native()` to both xls and xlsx paths on the R side.

PR #493, opened 2018-07-11
libxls gcc5/6/7 fix

breichholf:
Wasn't compiling because `ssize_t` was undefined.

```
gcc [...]   -fpic  -O2 -march=native -std=c99  -c endian.c -o endian.o
In file included from ./libxls/xlsstruct.h:40:0,
                 from libxls/endian.h:30,
                 from endian.c:33:
./libxls/ole.h:174:1: error: unknown type name ‘ssize_t’
 extern ssize_t ole2_read(void* buf,size_t size,size_t count,OLE2Stream* olest);
 ^
make: *** [endian.o] Error 1
ERROR: compilation failed for package ‘readxl’
```

`ssize_t` should be defined in `stdio.h`, according to [POSIX 2008](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdio.h.html#tag_13_49), which is included already and the compile flag `_XOPEN_SOURCE` was set to 700, but according to [POSIX 2004](http://pubs.opengroup.org/onlinepubs/9699919799/) and [POSIX 1995](http://pubs.opengroup.org/onlinepubs/007908799/) it is only defined in `sys/types.h`. Including that fixed the compile error.

If undefined, I further set `_XOPEN_SOURCE 500`, just in case.

### DIFF
--- a/src/libxls/ole.h
+++ b/src/libxls/ole.h
@@ -33,9 +33,12 @@
 #ifndef OLE_INCLUDE
 #define OLE_INCLUDE
 
-#define _XOPEN_SOURCE 500
-
 #include <stdio.h>			// FILE *
+
+#ifndef _XOPEN_SOURCE
+#define _XOPEN_SOURCE 500
+#endif
+#include <sys/types.h>
 
 #include "libxls/xlstypes.h"
 
--- a/src/libxls/ole.h
+++ b/src/libxls/ole.h
@@ -33,6 +33,8 @@
 #ifndef OLE_INCLUDE
 #define OLE_INCLUDE
 
+#define _XOPEN_SOURCE 500
+
 #include <stdio.h>			// FILE *
 
 #include "libxls/xlstypes.h"


jennybc:
This file is part of the embedded libxls, maintained here: https://github.com/evanmiller/libxls.

I am generally loathe to make any patches here that are not peculiar to R and readxl. I also don't see any problems around this on any of the platforms tested by CRAN, which makes me reluctant to act.

I suggest you bring it up on libxls itself.

breichholf:
Cheers and I agree. I read that it was pulled in from upstream, but didn't where from. I just found the original svn repo on sourceforge, ...
I'll open an issue there.

jennybc:
Yeah, I'd rather take the change from libxls itself. I pull from upstream periodically and, specifically, always before a readxl release.

jennybc:
I also point out that CRAN uses gcc 7 on many platforms and can build readxl with no problems. So this may be due to some other configuration issue on your system.

https://cran.r-project.org/web/checks/check_flavors.html#r-patched-linux-x86_64

breichholf:
After reading up more on `gcc` and why this error was happening I found a much simpler solution.
**Problem:** For some reason, R decides to pass `-std=c99` to gcc on my system, which results in the error described above, as `ssize_t` is not defined when called.
**Solution:** Add `CFLAGS=-std=gnu11` and `CXXFLAGS=-std=gnu11` to `~/.R/Makevars`, that then links POSIX libraries and everything compiles fine.

Issue #492, opened 2018-07-09
Multiple header rows mini-vignette?


apreshill:
Hi Jenny,

I accept the challenge to contribute a mini-vignette 😄  Do you have a preference on dataset? Would you prefer a small toy example dataset as a demo (as opposed to the data I used in my [blog post](https://alison.rbind.io/post/read-multiple-header-rows/))?

Thanks!
Alison

jennybc:
I just added a second sheet to the clippy.xls[x] example that already ships with readxl. I think it's sufficiently annoying? See what you think. I'm happy to get suggestions for change or if you want me to add a couple of rows.

<img width="288" alt="screen shot 2018-07-10 at 1 44 29 pm" src="https://user-images.githubusercontent.com/599454/42488141-7d362756-8447-11e8-8ad2-7c0788eba45f.png">

jennybc:
I assume it will go something like this and here's how to access the example sheet:

``` r
library(magrittr)
library(readxl)

(cnames <- readxl_example("clippy.xlsx") %>% 
    read_excel(sheet = 2, n_max = 0) %>% 
    names())
#> [1] "name"    "species" "death"   "weight"

readxl_example("clippy.xlsx") %>% 
  read_excel(sheet = 2, skip = 2, col_names = cnames)
#> # A tibble: 1 x 4
#>   name   species   death               weight
#>   <chr>  <chr>     <dttm>               <dbl>
#> 1 Clippy paperclip 2007-01-01 00:00:00    0.9
```

Created on 2018-07-10 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0.9000).

brianwdavis:
Thanks for taking this on @apreshill! 

If you look at the issue I opened (#486), I often have files where the first row isn't enough header to uniquely identify a variable, and we often need the data dictionary to be kept with the name. Following the `clippy.xls` example, I suppose I'd want to have the column `"death"/"(date is approximate)"` changed to `"date"/"of death"` and another column added `"date"/"of birth"` (with value 1/1/1997). However I can also see where this would add unnecessary complexity.

My suggestion if you want to add that as a use case is to move the metadata extraction up to Step 1, and then `full_cnames <- clippy_meta %>% lapply(paste, sep = "_") %>% unlist()` and use that as an alternative option for the names in Step 2.

apreshill:
Hi @brianwdavis and @jennybc -

[...]

PR #494, opened 2018-07-11
add mini-vignette on multiple header rows

apreshill:
Hi @jennybc,

Here is a draft of the mini-vignette we talked about in issue #492. Happy to make any edits requested!

Thanks,
Alison


man/excel_format.Rd
```diff
@@ -29,7 +29,7 @@ Determine if files are xls or xlsx (or from the xlsx family).
 determine format. It draws on logic from two lower level functions:
 \itemize{
 \item \code{format_from_ext()} attempts to determine format from the file extension.
-\item \code{format_from_signature()} consults the \href{https://en.wikipedia.org/wiki/List_of_file_signatures}{file signature} or "magic
+\item \code{format_from_signature()} consults the \href{https://en.wikipedia.org/wiki/List_of_file_signatures}{filesignature} or "magic
```
jennybc: This off-target change comes from running `document()` and having not-the-latest version of roxygen2 (here's a whitespace bug in CRAN roxygen that's fixed in dev). You can add a commit that puts this file back in its original state.

vignettes/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.html
+*.R
```
jennybc: Is there any specific reason to add this .gitignore file?

man/excel_format.Rd
apreshill: will do!

vignettes/.gitignore
apreshill: no! I believe it was inserted totally inadvertently, will revert in my branch

apreshill:
Please let me know if this fixes things! It looks like that errant `git.ignore` file was added because I used `usethis::use_vignette("multiple-header-rows")` (I think!).

jennybc:
Sorry, one last thing. Will you add a bullet to NEWS.md? A similar thing happened reprex, so you could do sg similar: https://github.com/tidyverse/reprex/blob/master/NEWS.md

apreshill:
Got it- I also followed Nick's example and added a short pointer at the end of the vignette, hope that is ok!

jennybc:
Thanks!

PR #501, opened 2018-08-21
Adapt to tibble v1.4.99.* and ruminate on name repair

jennybc:
For discussion. This PR

1. Adapts to the deprecation of `validate` in `as_tibble()`, i.e. it uses the new `.tidy_names` argument.
1. Intercepts `.tidy_names = TRUE` and `.tidy_names = some_function`, in order to capture the names before and after repair and return them in the `names_history` attribute of the returned tibble. This is in the same spirit as the "problems" reported by readr.

Reminder of what `.tidy_names` does:

https://github.com/tidyverse/tibble/blob/e816ba30ad227f7018a1868a66cc324d53589ffd/R/tibble.R#L26-L31

```
#' @param .tidy_names Treatment of invalid or duplicate column names:
#'   - `NULL`: default, throw an error if there are any missing or duplicated names,
#'   - `FALSE`: deliberately request a tibble with invalid names,
#'   - `TRUE`: apply [tidy_names()] to the names,
#'   - a function: apply custom name repair (e.g., `.tidy_names = make.names`
#'     to get base R equivalence).
```

This adds quite a lot of code and requires adding rlang to Imports (or copying code from tibble and rlang into readxl). There are some changes that could be made in tibble to make this easier (e.g. https://github.com/tidyverse/tibble/issues/462).

We could also decide to decline to report on name repair. If user wants such info, they should prevent name repair via `read_excel(..., .tidy_names = FALSE)` and do it themselves after the fact.

(I haven't fixed tests or NEWS yet ... but this does work, usage below.)

---

``` r
devtools::load_all(".")
#> Loading readxl

path <- test_sheet("unnamed-duplicated-columns.xlsx")

read_excel(path)
#> New names:
#> * `` -> ..1
#> * var2 -> var2..2
#> * `` -> ..3
#> * var2 -> var2..4
#> # A tibble: 2 x 4
#>     ..1 var2..2   ..3 var2..4
#>   <dbl> <chr>   <dbl> <chr>  
#> 1     1 a         1.1 aa     
#> 2     2 b         2.1 bb
read_excel(path, .tidy_names = TRUE) # the default
#> New names:
#> * `` -> ..1
#> * var2 -> var2..2
#> * `` -> ..3
#> * var2 -> var2..4
#> # A tibble: 2 x 4
#>     ..1 var2..2   ..3 var2..4
#>   <dbl> <chr>   <dbl> <chr>  
#> 1     1 a         1.1 aa     
#> 2     2 b         2.1 bb

read_excel(path, .tidy_names = NULL)
#> Error: Columns 1, 3 must be named.

read_excel(path, .tidy_names = FALSE)
#> # A tibble: 2 x 4
#>      `` var2     `` var2 
#>   <dbl> <chr> <dbl> <chr>
#> 1     1 a       1.1 aa   
#> 2     2 b       2.1 bb

## base R name repair
read_excel(path, .tidy_names = make.names)
#> # A tibble: 2 x 4
#>       X var2      X var2 
#>   <dbl> <chr> <dbl> <chr>
#> 1     1 a       1.1 aa   
#> 2     2 b       2.1 bb

## emulating behaviour of readxl v1.1.0
f <- function(x) {
  x <- tibble::repair_names(x, prefix = "X", sep = "__")
  names(x)
}
read_excel(path, .tidy_names = f)
#> # A tibble: 2 x 4
#>    X__1 X__2   X__3 X__4 
#>   <dbl> <chr> <dbl> <chr>
#> 1     1 a       1.1 aa   
#> 2     2 b       2.1 bb


## now let's look at names before / after
x <- read_excel(path, .tidy_names = TRUE)
#> New names:
#> * `` -> ..1
#> * var2 -> var2..2
#> * `` -> ..3
#> * var2 -> var2..4
attr(x, "names_history")
#> # A tibble: 4 x 2
#>   before after  
#>   <chr>  <chr>  
#> 1 ""     ..1    
#> 2 var2   var2..2
#> 3 ""     ..3    
#> 4 var2   var2..4

x <- read_excel(path, .tidy_names = make.names)
attr(x, "names_history")
#> # A tibble: 4 x 2
#>   before after
#>   <chr>  <chr>
#> 1 ""     X    
#> 2 var2   var2 
#> 3 ""     X    
#> 4 var2   var2

x <- read_excel(path, .tidy_names = f)
attr(x, "names_history")
#> # A tibble: 4 x 2
#>   before after
#>   <chr>  <chr>
#> 1 ""     X__1 
#> 2 var2   X__2 
#> 3 ""     X__3 
#> 4 var2   X__4
```

Created on 2018-08-21 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0.9000).

### DIFF
--- a/R/read_excel.R
+++ b/R/read_excel.R
@@ -38,6 +38,7 @@ NULL
 #'   the returned tibble. Ignored if `range` is given.
 #' @param guess_max Maximum number of data rows to use for guessing column
 #'   types.
+#' @inheritParams tibble::tibble
 #' @return A [tibble][tibble::tibble-package]
 #' @seealso [cell-specification] for more details on targetting cells with the
 #'   `range` argument
@@ -84,7 +85,8 @@ NULL
 read_excel <- function(path, sheet = NULL, range = NULL,
                        col_names = TRUE, col_types = NULL,
                        na = "", trim_ws = TRUE, skip = 0, n_max = Inf,
-                       guess_max = min(1000, n_max)) {
+                       guess_max = min(1000, n_max),
+                       .tidy_names = TRUE) {
   path <- check_file(path)
   format <- check_format(path)
   read_excel_(
@@ -92,7 +94,7 @@ read_excel <- function(path, sheet = NULL, range = NULL,
     col_names = col_names, col_types = col_types,
     na = na, trim_ws = trim_ws, skip = skip,
     n_max = n_max, guess_max = guess_max,
-    format = format
+    format = format, .tidy_names = .tidy_names
   )
 }
 
@@ -105,13 +107,14 @@ read_excel <- function(path, sheet = NULL, range = NULL,
 read_xls <- function(path, sheet = NULL, range = NULL,
                      col_names = TRUE, col_types = NULL,
                      na = "", trim_ws = TRUE, skip = 0, n_max = Inf,
-                     guess_max = min(1000, n_max)) {
+                     guess_max = min(1000, n_max),
+                     .tidy_names = TRUE) {
   path <- check_file(path)
   read_excel_(
     path = path, sheet = sheet, range = range,
     col_names = col_names, col_types = col_types,
     na = na, skip = skip, n_max = n_max, guess_max = guess_max,
-    format = "xls"
+    format = "xls", .tidy_names = .tidy_names
   )
 }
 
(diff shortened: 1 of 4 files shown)


jennybc:
cc @krlmlr

jennybc:
@hadley and I resolved while reviewing this that I will NOT worry about returning a tibble with name repair info in an attribute. The new control over name repair afforded by `.tidy_names` (or whatever) is more than sufficient to give people the control and info they need.

R/read_excel.R
```diff
@@ -149,17 +154,58 @@ read_excel_ <- function(path, sheet = NULL, range = NULL,
...
-      validate = FALSE
+
+  ret <- read_fun(
+    path = enc2native(normalizePath(path)), sheet_i = sheet,
+    limits = limits, shim = shim,
+    col_names = col_names, col_types = col_types,
+    na = na, trim_ws = trim_ws, guess_max = guess_max
+  )
+
+  if (utils::packageVersion("tibble") <= "1.4.2") {
```
krlmlr: Do you want to be explicit about this switch, to simplify later troubleshooting?

R/read_excel.R
```diff
@@ -149,17 +154,58 @@ read_excel_ <- function(path, sheet = NULL, range = NULL,
...
+    ## readxl v1.1.0 behaviour
+    return(
+      tibble::repair_names(
+        tibble::as_tibble(ret, validate = FALSE),
+        prefix = "X", sep = "__"
+      )
     )
+  }
+
+  ## re-implementing tibble:::auto_tidy_names()
```
krlmlr: Why do you need to reimplement? Isn't it sufficient to call `tidy_names()` ?

R/read_excel.R
```diff
@@ -149,17 +154,58 @@ read_excel_ <- function(path, sheet = NULL, range = NULL,
...
+  nms_before <- names(ret)
+
+  ## .tidy_names is either TRUE or a function
+  if (isTRUE(.tidy_names)) {
+    .tidy_names <- tibble::tidy_names
+  }
+
+  ## https://github.com/tidyverse/tibble/issues/462
+  ## poses questions about the use of rlang::names2()
+  if (rlang::is_function(.tidy_names)) {
```
jimhester: If these are the only places you are using `rlang`, I think I would consider avoiding the dependency and just provide a simple implementation of names2 in readxl.

hadley: Yeah, I think it's basically pointless to worry about an rlang dependency for tidyverse packages (but it's worth at least thinking about it for r-lib packages)

R/read_excel.R
jennybc: That was necessary as part of the experiment to return before/after names in an attribute.

PR #503, opened 2018-08-30
Change rapidxml link to regular http

### LINKED ISSUE #502: URL of rapidxml in DESCRIPTION not working
The URL of rapidxml in the DESCRIPTION is given as 'https://rapidxml.sourceforge.net' which does not work. 

Elsewhere in the readxl package rapidxml is referenced as 'http://rapidxml.sourceforge.net' which works perfectly.

batpigandme:
* Fixes #502
* If we want to avoid having non-https links in DESCRIPTION, we could just remove the link there, since we mention it elsewhere


jennybc:
Thanks!

PR #505, opened 2018-09-12
Accept XLSX files that do not contain a "styles" part

j6t:
An XLSX file that does not contain a "styles" part triggers this error:

  > read_excel("foo.xlsx", sheet="Sheet1")
  Error in sheets_fun(path) :
    Evaluation error: Couldn't find '' in 'foo.xlsx'.

To fix this, mimic the check that cacheStringTable() performs also in
cacheDateFormats().


jennybc:
Can you open an issue with a bit more background and an example of a file with this quality? I don't believe I've ever seen this phenomenon. We'd also need related tests.

j6t:
Thank you for your feedback. I've submitted issue #506 and added a test case to my proposed fix.

jennybc:
We work on things in cycles and readxl is not in such a phase right now. But I'll try to take a look to get you unstuck.

jennybc:
Thanks!

I made a few small changes. Usually I can push back into someone's PR but that did not work here. So I continued in an internal branch and merged. You are credited in the NEWS.

https://github.com/tidyverse/readxl/pull/528

jennybc:
In future and in general, it is best if you make a pull request from a non-`master` branch in your fork.

PR #517, opened 2018-11-13
Prepare for changes re: name repair in tibble

jennybc:
The new `.name_repair` looks good! Consider this a pilot test for what other downstream packages will need to do.

To support new tibble, I had to stop calling `as_tibble()` with the `validate` argument  and start passing `.name_repair`, with a readxl default value of "unique".

Vast majority of my changes are around messaging.

---

``` r
devtools::load_all()
#> Loading readxl

utils::packageVersion("tibble")
#> [1] '1.4.99.9005'
```

Working with a test sheet in which all col names are the empty string.

readxl default behavior = `.name_repair = “unique”`

``` r
read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names"
)
#> New names:
#> * `` -> `..1`
#> * `` -> `..2`
#> # A tibble: 1 x 2
#>   ..1   ..2  
#>   <chr> <chr>
#> 1 a     b
```

Demonstrating all the promised flexibility of `.repair_names` is now available here! First, using levels “minimal” and “universal”.

``` r
read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names",
  .name_repair = "minimal"
)
#> # A tibble: 1 x 2
#>   ``    ``   
#>   <chr> <chr>
#> 1 a     b

read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names",
  .name_repair = "universal"
)
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> # A tibble: 1 x 2
#>   ...1  ...2 
#>   <chr> <chr>
#> 1 a     b
```

Now I pass a custom namer.

``` r
LETTER_names <- function(nms) LETTERS[seq_along(nms)]
read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names",
  .name_repair = LETTER_names
)
#> # A tibble: 1 x 2
#>   A     B    
#>   <chr> <chr>
#> 1 a     b
```

Using a purrr-style formula to use `base::make.names()`

``` r
read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names",
  .name_repair = ~ make.names(., unique = TRUE)
)
#> # A tibble: 1 x 2
#>   X     X.1  
#>   <chr> <chr>
#> 1 a     b
```

<sup>Created on 2018-11-12 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


jennybc:
Here's usage with tibble v1.4.12 (current CRAN version) installed.

---

``` r
devtools::load_all()
#> Loading readxl

utils::packageVersion("tibble")
#> [1] '1.4.2'
```

Full message upon first call.

``` r
read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names",
  .name_repair = ~ make.names(., unique = TRUE)
)
#> readxl works best with a newer version of the tibble package.
#> You currently have tibble v1.4.2.
#> Falling back to column name repair from tibble <= v1.4.2.
#> New names:
#>  -> ..1
#>  -> ..2
#> # A tibble: 1 x 2
#>   ..1   ..2  
#>   <chr> <chr>
#> 1 a     b
```

Subsequently we only message if user seems to be explicitly trying to use `.name_repair`.

``` r
read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names"
)
#> New names:
#>  -> ..1
#>  -> ..2
#> # A tibble: 1 x 2
#>   ..1   ..2  
#>   <chr> <chr>
#> 1 a     b

read_excel(
  test_sheet("no-yes-col-names.xlsx"),
  sheet = "empty-col-names",
  .name_repair = ~ make.names(., unique = TRUE)
)
#> Update the tibble package to use the `.name_repair` argument. Ignoring.
#> New names:
#>  -> ..1
#>  -> ..2
#> # A tibble: 1 x 2
#>   ..1   ..2  
#>   <chr> <chr>
#> 1 a     b
```

<sup>Created on 2018-11-12 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

R/read_excel.R
```diff
@@ -38,6 +38,11 @@ NULL
 #'   the returned tibble. Ignored if `range` is given.
 #' @param guess_max Maximum number of data rows to use for guessing column
 #'   types.
+#' @param .name_repair Treatment of problematic column names. If the tibble
+#'   package version is recent enough, this offers full access to `.name_repair`
+#'   as documented in [tibble::tibble()]. Otherwise, this falls back to name
+#'   repair from tibble v1.4.2. Default behaviour ensures column names are not
+#'   empty and are unique.
```
jennybc: Ideally this would `@inheritParams tibble::tibble`, but that would make new tibble a hard requirement. Which is worth considering, actually

R/read_excel.R
```diff
@@ -84,14 +89,16 @@ NULL
 read_excel <- function(path, sheet = NULL, range = NULL,
                        col_names = TRUE, col_types = NULL,
                        na = "", trim_ws = TRUE, skip = 0, n_max = Inf,
-                       guess_max = min(1000, n_max)) {
+                       guess_max = min(1000, n_max),
+                       .name_repair = "unique") {
```
jennybc: I think this has to be the default here in readxl (versus "check_unique" in tibble), for backwards compatibility. readxl has ensured names are unique for a couple versions now.

DESCRIPTION
```diff
@@ -46,7 +46,8 @@ BugReports: https://github.com/tidyverse/readxl/issues
 Imports:
     cellranger,
     Rcpp (>= 0.12.12),
-    tibble (>= 1.3.1)
+    tibble (>= 1.3.1),
```
krlmlr: We can do at least >= 1.4.2 now.

jennybc:
Well, I would call this the *deluxe* treatment, because I'm actually exposing `.name_repair` and being scrupulous about messaging.

Feels like we should experiment with a few other packages to get a better sample before deciding. `validate` is actually ok to use, but it provokes a message. We could turn that off or make it "once per session".

R/read_excel.R
hadley: Is there a good reason why we shouldn't require the latest version of tibble?

jennybc: Not really, no.

Practically, there's the question of sequencing the CRAN submissions. And a user's possible irritation that they *must* do X right now, if they want to do Y.

jennybc:
Note to self re: backwards-compatibility.

The backwards compatibility target should be readxl v1.1.0 behaviour, which (I think) is:

```
tibble::repair_names(
  tibble::as_tibble(ret, validate = FALSE),
  prefix = "X", sep = "__"
)
```

Current master is doing something "in between". It is neither acting like readxl v1.1.0 nor is it fully caught up with dev tibble.

PR #523, opened 2018-12-01
Move C_VISIBILITY out of CPPFLAGS

jeroen:
as requested by CRAN.


jennybc:
Thanks!

How do I reconcile your resolution, which now has:

```
PKG_CXXFLAGS = $(C_VISIBILITY)
```

with what Prof Ripley says:

> Visibility flags are compiler flags not preprocessor flags, so 
`$(C_VISIBILITY)` should rather be in `PKG_CFLAGS` .  (And `C_VISIBILITY` is 
not appropriate for the C++ compiler ...).

?

I'd like to keep the flag, in whatever forms are most appropriate because, for example, another package could conceivably bundle RapidXML.

jeroen:
Ah okay now BDR's comment makes sense. However `CXX_VISIBILITY` is not available for old R versions so there is no easy way to use this....

wch:
Will it cause problems to try to use `CXX_VISIBILITY` in an older R version? I'm using R 3.5.1, and I have `PKG_CXXFLAGS = $(CXX_VISIBILITY)`, and httpuv compiles and runs just fine.

jeroen:
It won't cause problems because it doesn't do anything. So you're also not protected form symbol collision problems.

jeroen:
So far we have only seen symbol collision issues with rstudio-server on CentOS and Fedora systems. Never on Ubuntu or MacOS or Windows. Here are the relevant issues:
 - hunspell: https://github.com/ropensci/hunspell/issues/21
 - commonmark: https://github.com/jeroen/commonmark/issues/12

wch:
For future reference: When we tested building httpuv on Mac with `PKG_CFLAGS` and `PKG_CXXFLAGS` set explicitly to `-fvisibility=hidden`, it appears that the functions can't be called from R. This is probably why `C_VISIBILITY` is not set in the Mac Makeconf file. 


```
> httpuv::startServer( ... )
 Error in .Call("_httpuv_makeTcpServer", PACKAGE = "httpuv", host, port,  : 
  "_httpuv_makeTcpServer" not available for .Call() for package "httpuv"
```

However, when I tested the same thing on Ubuntu Linux, it worked fine. (On that platform, the Makeconf file has `C_VISIBILITY=-fvisibility=hidden`.)

At any rate, it looks like things work fine for httpuv with:

```
PKG_CFLAGS = $(C_VISIBILITY)
PKG_CXXFLAGS = $(CXX_VISIBILITY)
```

PR #538, opened 2018-12-18
Indicate progress via spinner

### LINKED ISSUE #243: read_excel progress bar
It would be nice to have a progress bar like readr.

Here's the "show_progress" function from https://github.com/tidyverse/readr/blob/8d6a892b941948c96b6a5aa2ddb5b897492478e3/R/utils.R

show_progress <- function() {
  !isTRUE(getOption("readr.show_progress")) || # user disables progress bar
  !interactive() || # not an interactive session
  !is.null(getOption("knitr.in.progress")) # Not actively knitting a document
}

jennybc:
WIP

Closes #243 read_excel progress bar


src/XlsWorkSheet.h
```diff
@@ -114,10 +119,16 @@ class XlsWorkSheet {
...
+    Spinner spinner("Guessing column types :spin");
+
+    // count is for spinner and checking for interrupt
+    int count = 0;
     // base is row the data starts on **in the spreadsheet**
     int base = cells_.begin()->row() + has_col_names;
     while (xcell != cells_.end() && xcell->row() - base < guess_max) {
-      if ((xcell->row() - base + 1) % 1000 == 0) {
+      count++;
+      if (count % 100000 == 0) {
```
jimhester: Feel free to file this under useless, premature optimizations, but using a power of two makes modulus much faster.

Since `x modulo y = (x & (y − 1))`

It goes from a division operation (which takes ~ 20+ processor cycles) to a single bitwise and, which can be done much faster. Compilers perform this optimization automatically when you use a power of 2, compare the instructions for `mod2` to `mod` at https://godbolt.org/z/Buce-D. (Note the compiler is even avoiding doing a division even in the first case, instead doing a multiply, an add and an exclusive or)

jennybc: I LOVE IT. I have other changes to make anyway, so will do.

jennybc:
@gaborcsardi I wrapped the C++ progress bar from progress here to 1) dumb it down into a spinner and 2) put it under the control of an argument on the R side. Feel free to ignore but I'm also happy to hear any feedback.

src/XlsWorkSheet.h
```diff
@@ -114,10 +123,14 @@ class XlsWorkSheet {
...
     }
 
+    // count is for spinner and checking for interrupt
+    int count = 0;
     // base is row the data starts on **in the spreadsheet**
     int base = cells_.begin()->row() + has_col_names;
     while (xcell != cells_.end() && xcell->row() - base < guess_max) {
-      if ((xcell->row() - base + 1) % 1000 == 0) {
+      count++;
+      if (count % 131072 == 0) {
```
hadley: I had to verify that this was 2 ^ 17.

jimhester: Unfortunately the optimization [doesn't seem to happen](https://godbolt.org/z/6YfGoQ) if you use `(2^17)` directly (even in a constant), but maybe you could use a named constant with a comment? e.g.

```cpp
const int PROGRESS_TICK = 131072; // 2^17
```

PR #559, opened 2019-03-09
Don't misinterpret colors as dates

### LINKED ISSUE #388: The 'd' in [red] is interpreted as a date format
https://github.com/tidyverse/readxl/blob/0d9ad4f570f6580ff716e0e9ba5048447048e9f0/src/ColSpec.h#L145-L148

That was too tempting for [somebody](https://github.com/CAPITAmodel/CAPITA), who created this beauty (see `MedicareAndLevies!Q:Q` ):
* Original file: [CPS v17-09-12.xlsb.zip](https://github.com/tidyverse/readxl/files/1327822/CPS.v17-09-12.xlsb.zip)
* Converted to xlsx using Excel 2010: [CPS v17-09-12.xlsx](https://github.com/tidyverse/readxl/files/1327821/CPS.v17-09-12.xlsx)

How to fix?

https://github.com/tidyverse/readxl/blob/0d9ad4f570f6580ff716e0e9ba5048447048e9f0/src/ColSpec.h#L150

... like the [awful hack](https://github.com/nacnudus/tidyxl/blob/89a9e0dec1ef74affcfda6124fb75d6d9707328e/src/styles.cpp#L254-L267) in tidyxl? :wink:

```cpp
// Adapted from hadley/readxl
bool styles::isDateFormat(std::string formatCode) {
  for (size_t i = 0; i < formatCode.size(); ++i) {
    switch (formatCode[i]) {
      case 'd':
        // Might be as in "[Red]"
        if (i < formatCode.size() - 1) {
          // there's at least one more character
          if (formatCode[i+1] == ']') {
            break;
          } else {
            return true;
          }
        }
```

Reprex:

``` r
library(tidyverse)
#> Loading tidyverse: ggplot2
#> Loading tidyverse: tibble
#> Loading tidyverse: tidyr
#> Loading tidyverse: readr
#> Loading tidyverse: purrr
#> Loading tidyverse: dplyr
#> Conflicts with tidy packages ----------------------------------------------
#> filter(): dplyr, stats
#> lag():    dplyr, stats
library(readxl)

source_url <- "https://github.com/tidyverse/readxl/files/1327821/CPS.v17-09-12.xlsx"
filename <- "CPS v17-09-12.xlsx"
sheetname <- "MedicareAndLevies"

download.file(source_url, filename, mode = "wb")

x <- read_xlsx(filename, sheetname, skip = 6)
class(x$MedLevSurRate2)
#> [1] "POSIXct" "POSIXt"
```

The culpable number formats:

```r
library(tidyxl)
y <- tidy_xlsx(filename, sheetname) # Slow. I wish I could call Jim Hester for advice.

numFmt <- tibble(numFmt = y$formats$local$numFmt,
                 id = seq_along(y$formats$local$numFmt))

y$data[[1]] %>%
  filter(row >= 8, col == 17) %>%
  left_join(numFmt, by = c("local_format_id" = "id")) %>%
  distinct(numeric, numFmt)
#> # A tibble: 3 x 2
#>   numeric                               numFmt
#>     <dbl>                                <chr>
#> 1  0.0100     "#,##0.00_ ;[Red]\\-#,##0.00\\ "
#> 2  0.0125 "#,##0.0000_ ;[Red]\\-#,##0.0000\\ "
#> 3      NA                              General
```

Kudos to @HughParsonage for discovering this file and daring to import it into R.

nacnudus:
Fixes #388
Closes #692

This caught me again recently.

Excel number formats support colours for, e.g. negative values `0.00;0.00[Red]`. Currently the `d` in `Red` (and some letters of other colours) is confused with the `d` in a date format `dd/mm/yyyy`, so the cell is incorrectly treated as a `date` cell.

The tests force a warning if this happens, by explicitly setting `col_type = "numeric"`.

The 'lowercase' variant of the test spreadsheet uses `red` (and similar) rather than `Red`, because this is allowed by the spec, though Excel itself doesn't create such files.


jennybc:
Thanks for tacking this!

Do we know if similar code is already in use somewhere and is therefore a bit battle-tested?

It would be very reassuring to use a variant of something that is proven to work.

nacnudus:
Hmm, this does deserve more work.  I hadn't really set aside time for it so I'll be a bit slow coming back to it.  Tidyxl has used this code for a while, but that's hardly battle tested so I'm comparing it with [xlrd](https://github.com/python-excel/xlrd/blob/15acaef790cba58a8e344b4e8691af8e014919c3/xlrd/formatting.py#L459), which takes a similar approach.  There are subtle bugs in both.

jennybc:
It doesn't have to be perfect -- we have a rather large bug at the moment. But it's _our_ bug and it seems to cause very few problems and I would hate to implement a solution that caused a net increase in user pain.

kcrt:
In the Chinese, Japanese, and Korean versions of Excel, the default representation of "number" is "0_);[Red]\(0\)".
So, many files are difficult to load because of this problem.
I hope this problem will be solved someday.

PR #581, opened 2019-08-21
List and filter sheets

aswansyahputra:
Hi,

I propose a functionality to filter excel sheets by using wild card or regular expression. For example:

```r
excel_sheets(readxl_example("datasets.xlsx"))
#> "iris"     "mtcars"   "chickwts" "quakes"  
excel_sheets(readxl_example("datasets.xlsx"), glob = "??i*")
#> "iris"     "chickwts"
excel_sheets(readxl_example("datasets.xlsx"), regexp = "(r|e)s$")
#> "mtcars" "quakes"
```

The purpose is to make it easier for user to pick certain sheets to be read simultaneously using `lapply()` or `map()`. Any suggestion is much appreciated!

Regards,

Aswan

### DIFF
--- a/R/excel-sheets.R
+++ b/R/excel-sheets.R
@@ -1,20 +1,49 @@
-#' List all sheets in an excel spreadsheet
+#' List sheets in an excel spreadsheet
 #'
 #' @inheritParams read_excel
+#' @param glob A wildcard to filter sheets.
+#' @param regexp A regular expression to filter sheets.
 #' @export
 #' @examples
 #' excel_sheets(readxl_example("datasets.xlsx"))
 #' excel_sheets(readxl_example("datasets.xls"))
 #'
+#' # Filter sheets using certain criteria
+#' excel_sheets(readxl_example("datasets.xls"), glob = "??i*")
+#' excel_sheets(readxl_example("datasets.xls"), regexp = "(r|e)s$")
+#'
 #' # To load all sheets in a workbook, use lapply
 #' path <- readxl_example("datasets.xls")
 #' lapply(excel_sheets(path), read_excel, path = path)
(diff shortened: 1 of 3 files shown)


jennybc:
Thanks! I can't quite decide how I feel about this.

I'll articulate the side of me that resists it: this feels like a result that is fairly easy for the the user to get via composition already. I.e. use `excel_sheets()` to get the names, then (say) pipe that to a regex filtering call, then pipe that into an iterative machine calling `read_excel(sheet = ...)`. So, why add arguments to existing functions when you can achieve the same by composing simple things?

What do you think? Do you have specific compelling example that helps us talk about this?

Even if my view prevails, you could develop such an example into a docs contribution somewhere in the package (help, workflow vignette, mini-vignette).

aswansyahputra:
Thanks for you response!

I agree that we can post-process the output of `excel_sheets()` in an easy way. But, in my opinion `excel_sheets()` shares similar idea with `list.files()` or `fs::dir_ls()` so that adding a filtering arguments to it seems to be natural. What do you think?

jennybc:
I think there are probably individual people and problem domains where the need to filter worksheets programmatically comes up a lot. And I gather this is true for you. But I don't think it's a very common need and it's easy to address via function composition. Hence my resistance.

aswansyahputra:
Yes I think your reasoning is true. I will close this PR then.

I truly appreciate your feedback on this, thanks!

jennybc:
This would still be very welcome as a mini-vignette. I like to help capture these common patterns, without necessarily building new functions and arguments into readxl itself.

Example of a mini-vignette on dealing with multiple header rows: https://readxl.tidyverse.org/articles/articles/multiple-header-rows.html

PR #596, opened 2019-11-07
Update helper.R

fermumen:
Added testthat's test_path function as recommended in the comments.
I was using readxl as a base for a package I'm writing that has some similarities and stumbled upon that comment.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,7 +53,7 @@ Suggests:
     knitr,
     rmarkdown,
     rprojroot (>= 1.1),
-    testthat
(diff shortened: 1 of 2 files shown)


jimhester:
You can see the tags any given commit is included on in the GitHub UI, e.g.

Screen Shot 2019-11-07 at 10 46 51 AM

So in this case `testthat (>= 2.0.0)` seems to be what you are after.

fermumen:
Ok, I have added the 2.0.0 dependency for testthat into the description file.

jennybc:
Thanks!

PR #602, opened 2020-01-31
Add actions

boshek:
Adding base github action and tidy template for pkgdown. Custom modification to `readxl` to follow in a separate PR because this PR will actually trigger GitHub actions. 

@jimhester

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -30,3 +30,5 @@
 ^investigations$
 ^src/patch-to-libxls-upstream\.R$
 ^LICENSE\.md$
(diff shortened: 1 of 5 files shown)


boshek:
Actions pass over here. I think this is ready for review. It passes here:

https://github.com/boshek/readxl/pull/1

 Things that might be worth revisiting:

- The names in the actions interface are a bit long and therefore harder to read. Might be annoying.
- I am using tibble 1.4.2 from github rather that the CRAN archive. This was the only way I was able to make this work.

PR #653, opened 2021-07-23
Put dependencies into their own subdirectories

jimhester:
This should hopefully make it easier to keep straight what is readxl
code and what is dependency code


jimhester:
When looking at the compiled files I found it hard to keep track of which files were from dependencies. I generally think it is cleaner to put dependency libs into subdirectories.

jimhester:
Sorry about the check failures in the initial PR, they should be fixed now.

jimhester:
The macOS builds were failing due to an ABI conflict with the Rcpp version in the cache. I bumped the cache, which resolved the issue. As discussed R 3.2 is failing due to lack of the 'digest' package, so that failure can be ignored.

I think this is safe to merge.

jennybc:
Thanks!

PR #654, opened 2021-07-26
update testthat to edition 3

### LINKED ISSUE #650: Update to use 3rd edition of testthat
https://testthat.r-lib.org/articles/third-edition.html#activating

cc @sbearrows

sbearrows:
Let me know if I've forgotten something @jimhester!

fixes #650


jimhester (approved):
LGTM

jennybc:
I was thinking we would also remove all the `context()` calls in this PR. In a typical test file @sbearrows you currently see something like:

``` r
context("Sheet geometry")

# then the tests start
```

You can delete the `context()` lines and the empty line just below.

Until we do that, with test that 3rd edition, we'll see this:

```
Warning (test-coercion.R:1:1): (code run outside of `test_that()`)
`context()` was deprecated in the 3rd edition.
Backtrace:
 1. testthat::context("coercion") test-coercion.R:1:0
 2. testthat:::edition_deprecate(3, "context()")
 3. ```

sbearrows:
I've removed all the instances of `context()` but now that I'm looking more closely at the output from `test()`, there are other deprecated function calls such as `expect_is()`.

Do you want me to update these as well?

jennybc:
Yeah that would be great, if it looks doable. Feel free to ask for guidance here (by commenting on specific lines of your PR) or in Slack, if it's hard to hang the question off a specific line of code.

sbearrows:
Since `all` is deprecated in `expect_warning()` there were a few cases were `expect_warning()` wasn't capturing the warning messages because there was more than one. Since in those instances the generation of the warning message wasn't being tested, I've changed them to `suppressWarning()` to clean up the output of `test()` but let me know if you disagree with this!

jennybc (commented):
I pushed one commit showing the use of `expect_snapshot()`.

I think that's a better way to adapt to what test that 3e does re: multiple warnings.

You can apply that approach to the other instances of this.

jennybc:
Thanks! I restyled because you introduced a few non-tidyverse-style things, which also cleaned up some existing style departures.

PR #655, opened 2021-07-26
update workflows

### LINKED ISSUE #649: Update CI workflows to current status
`usethis::use_tidy_github_actions()`, then tweak as needed if there are customizations specific to readxl.

sbearrows:
Unsure if there are changes we need to make for the readxl workflow @jimhester 

fixes #649


.github/workflows/pkgdown.yaml
```diff
@@ -1,45 +1,61 @@
...
+        shell: Rscript {0}
 
       - name: Install dependencies
         run: |
-          install.packages("remotes")
-          remotes::install_deps(dependencies = TRUE)
-          remotes::install_dev("pkgdown")
-          remotes::install_github("tidyverse/tidytemplate")
-          remotes::install_cran("tidyverse")
-          remotes::install_cran("devtools")
```
jimhester, line 37: We need to add these packages to the `DESCRIPTION` file in the `Config/Needs/website` field. e.g.

```
Config/Needs/website: tidyverse/tidytemplate, tidyverse, devtools
```

.github/workflows/test-coverage.yaml
```diff
@@ -1,43 +1,56 @@
...
         run: |
-          install.packages(c("remotes"))
-          remotes::install_deps(dependencies = TRUE)
-          remotes::install_cran("covr")
+          pak::local_install_dev_deps(upgrade = TRUE)
+          pak::pkg_install("covr")
         shell: Rscript {0}
 
       - name: Test coverage
-        run: covr::codecov(line_exclusions = c("src/xls.c", "src/xlstool.c", "src/rapidxml.h", "src/rapidxml_print.h", "src/endian.c", "src/ole.c", "src/libxls/brdb.h"))
```
jimhester, line 42: Instead of specifying these files here I would suggest you use a `.covrignore` file. The contents are regular expressions, similar to `.Rbuildignore`. This would allow them to be picked up by covr locally, and avoids needing customization in the workflows.

The contents of `.covrignore` would be something like the following.

```
src/xls\.c
src/xlstool\.c
src/rapidxml.*
src/endian\.c
src/ole\.c
src/libxls/brdb\.h
```

README.Rmd
```diff
@@ -19,6 +19,7 @@ knitr::opts_chunk$set(
 [![R build status](https://github.com/tidyverse/readxl/workflows/R-CMD-check/badge.svg)](https://github.com/tidyverse/readxl/actions)
 [![Codecov test coverage](https://codecov.io/gh/tidyverse/readxl/branch/master/graph/badge.svg)](https://codecov.io/gh/tidyverse/readxl?branch=master)
 [![lifecycle](https://img.shields.io/badge/lifecycle-stable-brightgreen.svg)](https://www.tidyverse.org/lifecycle/#stable)
+[![R-CMD-check](https://github.com/tidyverse/readxl/workflows/R-CMD-check/badge.svg)](https://github.com/tidyverse/readxl/actions)
```
jimhester, line 22: You should delete line 19 and replace it with this line

jimhester (changes requested):
The workflows themselves don't need changes, but there are some other files we should add information to, I left comments inline

.github/workflows/pkgdown.yaml
jennybc: I was using the dev version of pkgdown for some reason, but until/unless we re-discover that need, we can let that go.

README.Rmd
jennybc: It would be good to also rebuild the README with `devtools::build_readme()`. Depending on how you made your local repo, there might be a git hook that actually "encourages" (= forces) you to do this.

jennybc:
I made one more commit to adapt this to the file layout adopted when we merged #653.

Thanks!

.covrignore
```diff
@@ -0,0 +1,10 @@
+src/xls.c
+src/xlstool.c
+src/rapidxml.h
+src/rapidxml_print.h
```
jimhester: These should not be here, they are interpreted as regular expressions (e.g. so the `.` will match any character) and are already covered by the cases further down.

jimhester: But I think with the current master we can just ignore the directory completely (which is what is done anyway), so this comment can be ignored, sorry for the noise 🙉

sbearrows: You're right, I had first started adding them by hand when I realized what you were asking for. I forgot to remove them before pushing. Thank you for the noise!👂🏻

PR #658, opened 2021-07-28
Update to libxls 1.6.2

### LINKED ISSUE #652: Update vendored libxls code
cc @jennybc, @sbearrows

jennybc:
Closes #652

TODO:

* [x] Reapply `cran.h` workaround ~(or perhaps declare it no longer necessary)~
* [x] Add libxls version to what we write in DESCRIPTION (not just SHA)
* [x] Check for any remaining up-applied traditional patches and apply or convince myself they are no longer relevant
* [x] While retaining existing history, return to starting state and re-apply all changes in the ideal order / changesets so we can clearly distinguish changes that came from libxls's development vs. our patches
* [x] Add NEWS bullet
* [x] Determine if AIX/Sun fix still needed https://github.com/tidyverse/readxl/pull/543/commits/ea4d9ae26af0d7de7b57440db09e9249ae184bb4. *From r-hub checks, I see no evidence I need addition iconv fixes. In fact, given that we say readxl doesn't work on Solaris, I'm not entirely sure why we ever had this 🤔*


src/libxls/xlstool.h
```diff
@@ -33,15 +33,14 @@
  *
  */
 
-#include "libxls/xlsstruct.h"
-/* Mask illegal functions for CMD check */
-#include "cran.h"
```
jimhester, line 38: It might be better to do this masking when you include 'xlstool.h' in the readxl files rather than here?

jennybc: I have not re-introduced the `"cran.h"` header yet, but I will definitely take this advice when I do so, which is basically one of the next tasks.

src/libxls/locale.c
```diff
@@ -33,7 +33,10 @@
 #include "libxls/locale.h"
 
 xls_locale_t xls_createlocale() {
-#if defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
+#if defined(__MINGW32__)
+    xls_locale_t loc;
+    return loc;
+#elif defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
     return _create_locale(LC_CTYPE, ".65001");
```
jimhester, line 40: _create_locale doesn't seem to be available on the older mingw64 or mingw32 runtimes.

So I did similar to https://github.com/msys2/MINGW-packages/commit/dba069ff0fe1c0121665035e68150f556e5fe788 and just return an un-initialized locale for these cases.

Not sure if there is a more satisfactory answer, and perhaps we should do this only conditionally on the Rtools35 toolchain...

jimhester:
Ok, everything seems to be passing now, let me know if there is anything else you need!

src/libxls/xlstool.h
jimhester: We might not actually need it anymore, it looks like all the cases of fprintf are behind `OLE_DEBUG` defines and R CMD check isn't complaining about it...

src/libxls/locale.c
jennybc: It does seem appealing to apply this workaround in the narrowest situation possible. Is there established way to detect the old toolchain in this preprocessor macro style, i.e. just add one more condition?

src/libxls/xlstool.h
jennybc: That would be nice!

src/libxls/locale.c
jimhester: Not too easily, but I think we could use something with `#defined(__MINGW32__)` and the GCC version.

src/libxls/xlstool.h
jennybc: I am still seeing the usual complaints.

```
N  checking compiled code ...
   File ‘readxl/libs/readxl.so’:
     Found ‘___stderrp’, possibly from ‘stderr’ (C)
       Objects: ‘libxls/ole.o’, ‘libxls/xls.o’
     Found ‘_printf’, possibly from ‘printf’ (C)
       Objects: ‘libxls/xlstool.o’, ‘libxls/ole.o’, ‘libxls/xls.o’
     Found ‘_putchar’, possibly from ‘putchar’ (C)
       Object: ‘libxls/xls.o’
     Found ‘_puts’, possibly from ‘printf’ (C), ‘puts’ (C)
       Objects: ‘libxls/xlstool.o’, ‘libxls/xls.o’
   
   Compiled code should not call entry points which might terminate R nor
   write to stdout/stderr instead of to the console, nor use Fortran I/O
   nor system RNGs.
```

jennybc: I think this is still necessary because the NOTE is about the compiled object files and has nothing to do with whether we are exercising this code (which we don't, I think).

src/libxls/locale.c
jeroen: You can use the `__MINGW64_VERSION_MAJOR` macro

PR #659, opened 2021-07-28
Convert readxl from Rcpp to cpp11

sbearrows:
This initial commit contains the first steps for [converting from Rcpp to cpp11](https://cpp11.r-lib.org/articles/converting.html#mechanics-of-converting-a-package-from-rcpp-1)

Now I'll convert file by file and push each time I've completed a file so we can review.


src/zip.cpp
```diff
@@ -1,20 +1,21 @@
...
+  return env;
 }
 
 std::string zip_buffer(const std::string& zip_path,
                        const std::string& file_path) {
-  Rcpp::Function zip_buffer = readxl("zip_buffer");
+  cpp11::function zip_buffer = readxl("zip_buffer");
 
-  Rcpp::RawVector xml = Rcpp::as<Rcpp::RawVector>(zip_buffer(zip_path, file_path));
+  cpp11::writable::raws xml = cpp11::as_cpp<cpp11::writable::raws>(zip_buffer(zip_path, file_path));
```
jimhester, line 18:
```suggestion
  cpp11::raws xml(zip_buffer(zip_path, file_path));
```

src/zip.cpp
```diff
@@ -41,10 +42,10 @@ std::string xml_print(std::string xml) {
...
 }
 
-// [[Rcpp::export]]
+[[cpp11::register]]
 void zip_xml(const std::string& zip_path,
              const std::string& file_path) {
 
   std::string buffer = zip_buffer(zip_path, file_path);
-  Rcout << xml_print(buffer);
+  Rcpp::Rcout << xml_print(buffer);
```
jimhester, line 50:
```suggestion
  Rprintf("%s", xml_print(buffer).c_str());
```

sbearrows:
I had to edit a few other files to define the macros `R_NO_REMAP` and `STRICT_R_HEADERS` to get rid of the error message:

`R headers were included before cpp11 headers and at least one of R_NO_REMAP or STRICT_R_HEADERS was not defined.`

It was easier then try to figure out how to reorganize the header files.

src/CellLimits.h
```diff
@@ -1,3 +1,6 @@
+#define R_NO_REMAP
+#define STRICT_R_HEADERS
```
jimhester, line 2: I think if you include `cpp11/R.hpp` you won't need to add these defines.

sbearrows: I tried to include `cpp11/R.hpp` but I don't think I had the headers included in the right order because I was still getting the error. I kind of gave up since defining these fixed my problem 👾

src/XlsxWorkSheet.cpp
```diff
@@ -14,30 +14,31 @@ IntegerVector parse_ref(std::string ref) {
 }
 
 [[cpp11::register]]
-List read_xlsx_(std::string path, int sheet_i,
-                IntegerVector limits, bool shim,
-                RObject col_names, RObject col_types,
+cpp11::list read_xlsx_(std::string path, int sheet_i,
+                cpp11::writable::doubles limits, bool shim,
```
jimhester, line 18: Do we need `writable` here?

src/XlsxWorkSheet.h
```diff
@@ -198,7 +204,7 @@ class XlsxWorkSheet {
 
       xcell->inferType(na, trimWs, wb_.stringTable(), dateFormats_);
       CellType type = xcell->type();
-      Rcpp::RObject column = cols[col];
+      cpp11::sexp column = cpp11::as_sexp(cols[col]);
```
jimhester, line 207: Do we need this call to `as_sexp()`? I would guess no? maybe just this?

```suggestion
      cpp11::sexp column(cols[col]);
```

src/XlsxWorkSheet.cpp
```diff
@@ -14,30 +14,31 @@ IntegerVector parse_ref(std::string ref) {
...
                 int guess_max = 1000, bool progress = true) {
 
   // Construct worksheet ----------------------------------------------
-  XlsxWorkSheet ws(path, sheet_i, limits, shim, progress);
+  XlsxWorkSheet ws(path, sheet_i, cpp11::as_integers(limits), shim, progress);
 
   // catches empty sheets and sheets where requested rectangle contains no data
   if (ws.nrow() == 0 && ws.ncol() == 0) {
-    return Rcpp::List(0);
+    cpp11::list x;
```
jimhester, line 28: This would need to be a writable list I think? Otherwise it is going to return `NULL` instead of `list()`
```suggestion
cpp11::writable::list();
```

src/XlsxWorkSheet.cpp
```diff
@@ -14,30 +14,31 @@ IntegerVector parse_ref(std::string ref) {
...
   }
 
   // Get column names -------------------------------------------------
-  CharacterVector colNames;
+  cpp11::writable::strings colNames;
   bool has_col_names = false;
   switch(TYPEOF(col_names)) {
   case STRSXP:
-    colNames = as<CharacterVector>(col_names);
+    colNames = cpp11::as_cpp<cpp11::strings>(col_names);
```
jimhester, line 37: I don't think you need any cast? just `colNames = col_names`

src/XlsxWorkSheet.cpp
```diff
@@ -47,7 +48,7 @@ List read_xlsx_(std::string path, int sheet_i,
   if (TYPEOF(col_types) != STRSXP) {
     Rcpp::stop("`col_types` must be a character vector");
   }
-  std::vector<ColType> colTypes = colTypeStrings(as<CharacterVector>(col_types));
+  std::vector<ColType> colTypes = colTypeStrings(cpp11::as_cpp<cpp11::strings>(col_types));
```
jimhester, line 51: Not sure you need a cast here either

jimhester (commented):
There were a few cases where as far as I could tell you were converting an R SEXP to another R SEXP using `as_cpp()`, which you should never need to do.

The mental model is `as_cpp()` is used to convert a SEXP to a native C++ type (like `std::string`, `int`, etc.) and `as_sexp()` is used to go the opposite way, e.g. convert a C++ type to an R SEXP.

But in general you should have to call these explicitly somewhat rarely, most of the time they can be done implicitly in the generated registration code.

src/XlsxWorkSheet.cpp
sbearrows: I'm using `cpp11::as_integers` and I thought that would count as being modified but I guess not!

src/XlsxWorkSheet.h
sbearrows: I ended up having to do something similar to what we did with
`std::vector<ColType> colTypes = colTypeStrings(static_cast<SEXP>(col_types));`
Because I was getting a constructor error but let me know if there is another solution!

src/ColSpec.h
```diff
@@ -228,6 +239,7 @@ inline Rcpp::CharacterVector reconcileNames(Rcpp::CharacterVector names,
 }
 
 inline Rcpp::RObject makeCol(ColType type, int n) {
+  //1234
```
jimhester, line 242: I think to fill these logicals we might need a little helper function

```
template <typename T, typename N>
T new_vector(R_xlen_t size, N val) {
  T out(size);
  std::fill(out.begin(), out.end(), val);
  return out;
}

/* which you could call like */
cpp11::writable::list({new_vector<cpp11::writable::doubles>(n, NA_REAL)});
```

src/CellLimits.h
```diff
@@ -1,11 +1,15 @@
...
 #define READXL_CELLLIMITS_
 
-#include <Rcpp.h>
+#include <cpp11/integers.hpp>
+#include <cpp11/doubles.hpp>
 #include "XlsCell.h"
 
 class CellLimits {
-  int minRow_, maxRow_, minCol_, maxCol_;
+  double minRow_, maxRow_, minCol_, maxCol_;
```
jennybc, line 12: I find this switch from `int` to `double` surprising (and the ones below) and presumably undesirable. What's the rationale for this?

src/StringSet.h
```diff
@@ -1,24 +1,30 @@
...
-#define READXL_STRINGSET_
+#define R_NO_REMAP
+#define STRICT_R_HEADERS
 
-#include <Rcpp.h>
+#include <cpp11/strings.hpp>
+#include <cstring>
 #include "utils.h"
+#include <set>
+#include <vector>
```
jennybc, line 8: Does it make more sense to keep the system includes, e.g. `<whatever>` together and above the ones defined here, e.g. `"utils.h"`?

src/XlsWorkSheet.cpp
```diff
@@ -1,15 +1,13 @@
...
 #include "libxls/xls.h"
-using namespace Rcpp;
 
-// [[Rcpp::export]]
-List read_xls_(std::string path, int sheet_i,
-               IntegerVector limits, bool shim,
-               RObject col_names, RObject col_types,
+[[cpp11::register]]
+cpp11::list read_xls_(std::string path, int sheet_i,
+               cpp11::doubles limits, bool shim,
```
jennybc, line 9: As mentioned previously, I feel weird about this switch from integer to double.

src/XlsWorkSheet.h
```diff
@@ -31,32 +37,32 @@ class XlsWorkSheet {
 public:
 
   XlsWorkSheet(const XlsWorkBook wb, int sheet_i,
-               Rcpp::IntegerVector limits, bool shim, bool progress):
+               cpp11::doubles limits, bool shim, bool progress):
```
jennybc, line 40: Ditto the concern re: integer --> double.

src/XlsxWorkSheet.cpp
```diff
@@ -1,22 +1,22 @@
...
+  return x;
 }
 
-// [[Rcpp::export]]
-List read_xlsx_(std::string path, int sheet_i,
-                IntegerVector limits, bool shim,
-                RObject col_names, RObject col_types,
+[[cpp11::register]]
+cpp11::list read_xlsx_(std::string path, int sheet_i,
+                cpp11::doubles limits, bool shim,
```
jennybc, line 18: Ditto the concern about these becoming doubles when they really are integer by nature.

src/XlsxWorkSheet.h
```diff
@@ -38,16 +44,16 @@ class XlsxWorkSheet {
 public:
 
   XlsxWorkSheet(const XlsxWorkBook wb, int sheet_i,
-                Rcpp::IntegerVector limits, bool shim, bool progress):
+                cpp11::doubles limits, bool shim, bool progress):
```
jennybc, line 47: The integer --> double thing.

tests/testthat/test-read-excel.R
```diff
@@ -20,6 +20,7 @@ test_that("xlsx is not read as xls and vice versa", {
     read_xls(test_sheet("iris-excel-xlsx.xlsx")),
     "libxls error: Unable to open file"
   )
+  #add a skip for now
```
jennybc, line 23: What does this refer to me? Meaning: I don't see any associated `skip()`.

src/CellLimits.h
sbearrows: It's because`standardise_limits()` returns its elements as `doubles`. Previously when using `Rcpp::IntegerVector` these were being coerced to `ints` but cpp11 does not automatically coerce `doubles` when using `cpp11::integers`.
I thought it would be more efficient to change this on the C side rather than try to convert each element in the named vector to integers though I admittedly don't use named vectors in R much.
If you disagree we can go the R route or if Jim is planning to release some of our work with cpp11 to CRAN soon, we did make a helper `cpp11::as_integers()` that we could use instead in the `XlsxWorkSheet.h` and `XlsWorkSheet.h` files.

src/StringSet.h
sbearrows: Yes! Thank you for reminding me about this.
I rearranged some of the header files to avoid conflicts with Rcpp and cpp11 libraries so I need to reorganize them (or organize them back to how you had them). I ended up using the macros `#define R_NO_REMAP` and `#define STRICT_R_HEADERS` instead to avoid the conflicts, which I'll remove since they won't be needed anymore once this PR is closed.

tests/testthat/test-read-excel.R
sbearrows: I think during a meeting, @jimhester and I we were trying to diagnose a crash that was happening and thought it might be one of these tests. Looks like we fixed it though so I'll remove this comment.

src/CellLimits.h
jennybc: I think a single `as.integer()` on the R side will address this, which feels more correct and also affects many fewer lines of code.

src/StringSet.h
```diff
@@ -1,24 +1,30 @@
...
+
+#ifndef READXL_STRINGSET_
+#define READXL_STRINGSET_
```
jimhester, line 11: These defines should be at the top of the file

src/ColSpec.h
```diff
@@ -1,9 +1,21 @@
...
+#include <cpp11/strings.hpp>
+#include <cpp11/doubles.hpp>
+#include <cpp11/integers.hpp>
+#include <cpp11/list.hpp>
+#include <cpp11/logicals.hpp>
+#include <cpp11/sexp.hpp>
+#include <cpp11/as.hpp>
 #include "StringSet.h"
+#include "libxls/xls.h"
+#include <Rcpp.h>
```
jimhester, line 18: Can get rid of Rcpp.h here

src/CellLimits.h
jimhester: @sbearrows we had done as @jennybc describes in https://github.com/tidyverse/readxl/pull/659/commits/3f7553f4599615b08f38490bd479ccb816fa9455. I think some other tests ended up failing due to this, but I agree it is worth revisiting.

jimhester (commented):
We need to remove the rest of the Rcpp references as well. You can find them with find my files or with `git grep Rcpp`

1. Remove Rcpp references from Imports and LinkingTo
2. Remove the importFrom Rcpp roxygen annotation in R/readxl.R (and run devtools::document() afterwards)
3. Remove any stray `#include "Rcpp.h"` directives

You may find some compilation errors, as Rcpp includes a number of standard headers which cpp11 may not. If any of these (like std::vector or std::map etc.) are used in the generated `cpp11.cpp` code you will need to include them in a special file, `src/readxl_types.h` which cpp11 will automatically include in the generated file.

jimhester:
It looks like the only test still failing is on windows release due to `std::assert` not being declared, you should be able to `#include <cassert>` before the file including the libxls headers to fix this.

src/CellLimits.h
sbearrows: Yes if we used `as.integer(limits)` it fails which is why I changed `CellLimits.h` instead. So instead I've done the following:
```
  names <- names(limits)
  limits <- as.integer(limits)
  names(limits) <- names
  limits
```
I tried other approaches but this was the only one I found that worked. This way, `CellLimits.h` and friends stay as `int`. It's not pretty and feels kind of circular, so let me know if there is a sleeker option.

R/read_excel.R
```diff
@@ -274,14 +274,18 @@ standardise_limits <- function(range, skip, n_max, has_col_names) {
...
     limits <- c(
-      min_row = limits[["ul"]][1] - 1,
-      max_row = limits[["lr"]][1] - 1,
-      min_col = limits[["ul"]][2] - 1,
-      max_col = limits[["lr"]][2] - 1
-    )
+      min_row = as.integer(limits[["ul"]][1] - 1),
+      max_row = as.integer(limits[["lr"]][1] - 1),
+      min_col = as.integer(limits[["ul"]][2] - 1),
+      max_col = as.integer(limits[["lr"]][2] - 1)
```
jimhester, line 280: I don't think you need this here since you are doing `limits <- as.integer(limits)` below?

sbearrows: Yep you're right. This was one of my other attempts to coerce to `int` that did not work!

R/read_excel.R
```diff
@@ -274,14 +273,18 @@ standardise_limits <- function(range, skip, n_max, has_col_names) {
...
-      min_row = limits[["ul"]][1] - 1,
-      max_row = limits[["lr"]][1] - 1,
-      min_col = limits[["ul"]][2] - 1,
-      max_col = limits[["lr"]][2] - 1
-    )
+      min_row = (limits[["ul"]][1] - 1),
+      max_row = (limits[["lr"]][1] - 1),
+      min_col = (limits[["ul"]][2] - 1),
+      max_col = (limits[["lr"]][2] - 1)
+      )
```
jennybc, line 280: Are the surrounding `()` leftover from an intermediate experiment?

tests/testthat/test-read-excel.R
```diff
@@ -95,7 +95,9 @@ test_that("trim_ws must be a logical", {
 
 ## xlsx: https://github.com/tidyverse/readxl/issues/370
 ## xls:  https://github.com/tidyverse/readxl/issues/476
+## TODO: revisit on windows os when conversion to using cpp11 is completed
```
jennybc, line 98: Is this one ready for revisitation?

src/StringSet.h
jennybc: The current state seems to be much more regular, but often puts the local headers before the system headers. It feels like a more conventional order would be: system, vendored code (libxls), written-by-us, which also reflects intended precedence. (Let me also acknowledge that this probably doesn't matter and is more a matter of style at this point.)

R/read_excel.R
sbearrows: Yes they are I'll remove them!

tests/testthat/test-read-excel.R
sbearrows: Yes I can remove that!

src/StringSet.h
sbearrows: Ah yes sorry! I should have reread your comment before executing based on my memory. I'll rearrange in that order.

jimhester:
Thanks a million!

jennybc:
Woohoo! 🎉

PR #662, opened 2021-07-30
Troubleshooting the problem with Windows + r-devel (or maybe just "older" Windows)

jennybc:
Relates to #661 

Historical re-enactment, where I will also be checking with win-builder and/or r-hub as I go


jennybc:
I abandoned this.

PR #665, opened 2021-08-13
add data import cheat sheet

averiperny:
Adds the data import cheat sheet to README.Rmd now that it includes readxl functions.


jennybc:
Thanks! I also did `devtools::build_readme()` here too.

Issue #498, opened 2018-08-20
xls path containing the "~" shortcut can't be read


ddheart:
Minor issue: the '~' character on macOS used to point to the home folder causes readxl to throw an error.
---

```r
library(readxl)
(xl <- read_excel('/Users/mbpro/codingprojects/tests/Book1.xls'))
#> New names:
#> abc -> abc..1
#> abc -> abc..2
#> # A tibble: 4 x 2
#>   abc..1 abc..2
#>    <dbl>  <dbl>
#> 1      4      2
#> 2      3      3
#> 3      7      5
#> 4      2      2
dput(xl)
#> structure(list(abc..1 = c(4, 3, 7, 2), abc..2 = c(2, 3, 5, 2)), row.names = c(NA, 
#> -4L), class = c("tbl_df", "tbl", "data.frame"))
read_excel('~/codingprojects/tests/Book1.xls')
#> Error in read_fun(path = enc2native(path), sheet_i = sheet, limits = limits, : Failed to open ~/codingprojects/tests/Book1.xls
```

jennybc:
Hmm ... a workaround in the meantime is:

`read_excel(normalizePath("~/path/to/your.xls"))`

jennybc:
I believe we have just rediscovered a good reason by xls paths were being normalized:

https://github.com/tidyverse/readxl/pull/477#issuecomment-387800926

PR #674, opened 2022-01-03
Push re-encoding of xls path into XlsWorkBook constructor

jennybc:
I'm making this PR to restore symmetry of `read_xls_()` and `read_xlsx_()`. I want their signature and body to be the same (for eventual de-deduplication). The symmetry was lost at the time of the Rcpp --> cpp11 conversion (33c79a58).

General recap of this path handling:

* User's path needs to be normalized to expand, e.g., a leading `~/`. At least for xls, we need to do this (see #498).
* As of R 3.5, `normalizePath()` (`path.expand()`, really) returns UTF-8 encoded output, even if the input was not UTF-8.
* Once we adopted cpp11, it also does automatic conversion to UTF-8 for strings.
* This is fine for xlsx because we ultimately open the file with R (see `zip_buffer()` in [`R/xlsx-zip.R`](https://github.com/tidyverse/readxl/blob/d8f789e173b42cb213b8da14979c11aa300eb38e/R/xlsx-zip.R)).
* For xls, we need to re-encode to the native encoding before we pass the path to libxls, which [uses `fopen()`](https://github.com/tidyverse/readxl/blob/d8f789e173b42cb213b8da14979c11aa300eb38e/src/libxls/ole.c#L625).

Relevant previous history:

* https://github.com/tidyverse/readxl/pull/477
* https://github.com/tidyverse/readxl/pull/499

### DIFF
--- a/R/excel-sheets.R
+++ b/R/excel-sheets.R
@@ -18,7 +18,7 @@
 excel_sheets <- function(path) {
   path <- check_file(path)
   format <- check_format(path)
-  path <- normalize_path(path)
+  path <- normalizePath(path)
 
   switch(format,
     xls = xls_sheets(path),
--- a/R/read_excel.R
+++ b/R/read_excel.R
@@ -194,7 +194,7 @@ read_excel_ <- function(path, sheet = NULL, range = NULL,
     sheets_fun <- xlsx_sheets
     read_fun <- read_xlsx_
   }
-  path <- normalize_path(path)
+  path <- normalizePath(path)
   sheet <- standardise_sheet(sheet, range, sheets_fun(path))
   shim <- !is.null(range)
   limits <- standardise_limits(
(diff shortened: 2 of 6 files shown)


jennybc:
@gaborcsardi Will you just provide a quick look? I've tried to summarize how we got here and I'm quite confident this is reasonable.

gaborcsardi:
> For xls, we need to re-encode to the native encoding before we pass the path to libxls, which uses fopen().

FWIW this will not work on Unicode paths if they cannot be represented in the single-byte native encoding, and `_wfopen()` should be used instead on Windows. But you probably already know this, and I understand that you might not want to fix this in this PR.

Looks good otherwise.

jennybc:
> But you probably already know this, and I understand that you might not want to fix this in this PR.

I consider this a libxls problem. I guess I could open an issue or even a PR there, but I think I've got higher priority stuff to work on in readxl.

PR #680, opened 2022-02-08
prefer char to array in find_*_of() where possible

MichaelChirico:
https://clang.llvm.org/extra/clang-tidy/checks/performance-faster-string-find.html


jennybc:
Good idea.

PR #683, opened 2022-02-18
Use a template class to remove xls/xlsx duplication 

### LINKED ISSUE #284: Unified cell/worksheet classes or generic functions

jennybc:
Closes #284


src/CellLimits.h
```diff
@@ -101,4 +102,74 @@ class CellLimits {
...
+
+  // if nominal min row or col is (funny) less than actual,
+  // add a shim cell to the front of cells and update actual
+  bool   shim_up = funny_lt(nominal.minRow(), actual.minRow());
+  bool shim_left = funny_lt(nominal.minCol(), actual.minCol());
+  if (shim_up || shim_left) {
+    int ul_row = funny_min(nominal.minRow(), actual.minRow());
+    int ul_col = funny_min(nominal.minCol(), actual.minCol());
+    T ul_shim(std::make_pair(ul_row, ul_col));
+    cells.insert(cells.begin(), ul_shim);
```
jennybc, line 149: Abstract performance concern: presumably this insertion causes the entire `cells` vector to be copied, which would seem undesirable. That being said, I don't think this happens very often and has never been raised as a concrete performance concern.

This code is not new in this PR. It is newly located in a templated function as opposed to a duplicated member function.

src/Sheet.h
```diff
@@ -0,0 +1,292 @@
...
+};
+
+template <typename T>
+class Sheet {
+
+  typename T::Book wb_;
+  typename T::CellSet cs_;
+
+public:
+  Sheet(const std::string& path,
```
jennybc, line 38: The responsibilities of the previous `Xls[x]WorkSheet` class have been divided across the new `Sheet` template class and new `Xls[x]CellSet` class:

*I'm not very satisfied with the `Sheet` name for this class.*

* `Xls[x]CellSet`: A slimmed down version of the previous `Xls[x]WorkSheet`. This class gains access to a specific worksheet, inside a specific workbook, and slurps the cells out of it, yielding a "bunch of cells", of either the xls or xlsx variety. Basically holds all of the workbook/worksheet logic that is specific to xls vs xlsx.
* `Sheet`: Coordinates the workbook and `Xls[x]CellSet` and uses the "bunch of cells" to create a list to return to R (where it is then made into a tibble). Given a "bunch of cells", main functions:
  - `colNames()`: Read the first row’s worth of cells into a character vector of column names.
  - `colTypes()`: Read the first `guess_max` row's worth of cells and determine the most suitable type for each (not-yet-specified) column.
  - `readCols()`: Read an entire “bunch of cells” into R vectors of specific types with specific names.

src/XlsCell.h
```diff
@@ -53,7 +53,8 @@ class XlsCell {
 
   void inferType(const StringSet& na,
                  const bool trimWs,
-                 const std::set<int>& dateFormats) {
+                 const std::set<int>& dateFormats,
+                 const std::vector<std::string>& stringTable) {
```
jennybc, line 57: Even though the shared string table is completely abstracted away by libxls, I added a stub for it to `XlsWorkBook` and it newly appears in the signature of the `XlsCell` functions `inferType()`, `asStdString()`, and `asCharSxp()`. This was all in the name of removing one slight difference in the public interface of xls and xlsx, thus clearing the way for the creation of the `Sheet<T>` template class.

src/XlsCellSet.h
```diff
@@ -0,0 +1,169 @@
...
+#include "cpp11/sexp.hpp"
+#include "cpp11/strings.hpp"
+
+class XlsCellSet {
+  // xls specifics
+  xls::xlsWorkBook* pWB_;
+  xls::xlsWorkSheet* pWS_;
+
+  // common to xls[x]
+  Spinner spinner_;
```
jennybc, line 24: Not currently wired up and, presumably, moving up into the `Sheet<T>` class (?).

src/XlsxCell.h
```diff
@@ -55,8 +55,8 @@ class XlsxCell {
 
   void inferType(const StringSet& na,
                  const bool trimWs,
-                 const std::vector<std::string>& stringTable,
-                 const std::set<int>& dateFormats) {
+                 const std::set<int>& dateFormats,
+                 const std::vector<std::string>& stringTable) {
```
jennybc, line 59: I changed the order of these due to vague hopes that the `stringTable` argument could eventually be made optional, given that it's needed for xlsx but not xls.

src/XlsxCellSet.h
```diff
@@ -0,0 +1,163 @@
...
+    // actual_ is updated to reflect the insertions.
+    if (shim) insertShims(cells_, nominal_, actual_);
+
+    nrow_ = (actual_.minRow() < 0) ? 0 : actual_.maxRow() - actual_.minRow() + 1;
+    ncol_ = (actual_.minCol() < 0) ? 0 : actual_.maxCol() - actual_.minCol() + 1;
+  }
+
+  ~XlsxCellSet() {
+    //Rprintf("XlsxCellSet() destructor\n");
+    // should I be doing something here?
```
jennybc, line 92: I'd like to discuss this.

tests/testthat/test-dates.R
```diff
@@ -20,10 +20,10 @@ test_that("date subsecond rounding works", {
   expect_identical(as.character(df$dttm), df$dttm_string)
 })
 
-## Lotus 1-2-3 leap year bug
+## Lotus 1-2-3 leap year bug ----
 ## #264, #148
-test_that("we get correct dates prior to March 1, 1900, in 1900 date system", {
-  ## xlsx
+test_that("we get correct dates prior to March 1, 1900, in 1900 date system [xlsx]", {
+  skip("Still need to figure this out")
```
jennybc, line 26: Locally, this is the only test failure I see.

Now that I've pushed, though, I see more signs of trouble from CI on Windows and Unbunt, i.e. segfaults. I do have a local branch, where I've added some debugging code to troubleshoot this test, that also segfaults. Yay? So hopefully I can run this down locally.

src/XlsxCellSet.h
```diff
@@ -0,0 +1,163 @@
...
+    rapidxml::xml_node<>* rootNode;
+
+    if (sheet_i >= wb.n_sheets()) {
+      cpp11::stop("Can't retrieve sheet in position %d, only %d sheet(s) found.",
+                  sheet_i + 1,  wb.n_sheets());
+    }
+    sheetName_ = wb.sheets()[sheet_i];
+    std::string sheetPath = wb.sheetPath(sheet_i);
+    spinner_.spin();
+    std::string sheet_ = zip_buffer(wb.path(), sheetPath);
```
hadley, line 54: This needs to belong to the class again but it's the memory that backs `sheetXml_`

hadley (approved):

PR #687, opened 2022-03-15
Fix compilation on Alpine Linux

gaborcsardi:
and other systems that use musl.

The problem is that the internal musl header that defines
`uid_t` and `gid_t` only defines it conditionally, if a macro is set.
This macro is set if the header is included from `unistd.h`,
but not if included from another header.

readxl uses `#pragma once`, so the header will not be included
twice, so `uid_t` might not be defined, if unistd.h is not
included early on.

The same story happens with `struct timeval` and `sys/time.h`.

Closes https://github.com/r-hub/r-minimal/issues/40

### DIFF
--- a/src/SheetView.h
+++ b/src/SheetView.h
@@ -1,6 +1,3 @@
-#include <unistd.h>
-#include <sys/time.h>
-
 #include "CellLimits.h"
 #include "ColSpec.h"
 #include "XlsWorkBook.h"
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks for the explanation!

bastistician:
Unfortunately, 9884b53341e475ec984ebe2ff1182ff1347d06ad broke Gabor's fix. Compilation on Alpine Linux fails

> if unistd.h is not included early on

and I think the "early on" is crucial here (same with `sys/time.h`). Thanks!

bastistician:
> Do you want to make a PR?

I don't plan to. All credit goes to Gábor for debugging this issue. :)

PR #692, opened 2022-03-23
DRAFT: More sophisticated detection of datetime-y formats

jennybc:
This is just to force CI.
Assuming all is well, I will extend / update https://github.com/tidyverse/readxl/pull/559 and then merge it, instead of this.

---

Adapted from @reviewher

Code has been used in @nacnudus's tidyxl for a long time

Also includes a refinement from @da11an


src/ColSpec.h
```diff
@@ -178,6 +178,47 @@ inline bool isDateFormat(std::string x) {
   return false;
 }
 
+// Adapted from @reviewher https://github.com/tidyverse/readxl/issues/388
+// See also ECMA Part 1 page 1785 (actual page 1795) section 18.8.31 "numFmts
+// (Number Formats)"
+#define CASEI(c) case c: case (c | 0x20)
+#define CMPLC(j,n) if(x[i+j] | (0x20 == n))
+inline bool isDateFormat2(std::string x) {
```
hadley, line 186: Might be worth exporting this function so you can test independently

PR #694, opened 2022-03-25
Rework libxls locale patch

jennybc:
readxl had a pre-existing patch related to R's Windows toolchain(s), around some locale-handling code introduced in libxls 1.6.2.

Patch was originally applied before Rtools 4.2 and UCRT was on the horizon. Most recently touched in 1430c78a6286851d4b24459cdb9c32f1b12e1bb0.

Recent `check_win_devel()` and CRAN submissions revealed that the patch needed a refresh for R-devel (will become 4.2), though the relevant tests only fail when the system language is set to German or, presumably, other non-English languages.

Main things to note re: where this PR ended up:

* Current patch is applied as narrowly as possible, i.e. only when we detect readxl is being compiled with Rtools 3.5.

* I fixed what I suspect was a pre-existing infelicity, where a set & capture / restore sequence seemed to have a locale category mismatch, i.e. `LC_CTYPE` on the way in, but `LC_ALL` on the way out.

### DIFF
--- a/src/libxls/locale.c
+++ b/src/libxls/locale.c
@@ -34,15 +34,17 @@
 
 #if defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
 #ifdef __GNUC__
-  // g++ 10 indicates Rtools42 and UCRT
-  #if __GNUC__ < 10
-    #define WINDOWS_BEFORE_RTOOLS_42
-  #endif
+    // Rtools42 | R 4.2 | GCC 10 | UCRT
+    // Rtools40 | R 4.0.x to 4.1.x | GCC 8.3.0 | MSVCRT
+    // Rtools35 | R 3.3.x to 3.6.x | GCC 4.9.3 | MSVCRT
+    #if __GNUC__ < 10
+      #define WINDOWS_BEFORE_RTOOLS_42
+    #endif
 #endif
 #endif
 
 #ifdef WINDOWS_BEFORE_RTOOLS_42
-static char* old_locale;
+    static char* old_locale;
 #endif
 
 xls_locale_t xls_createlocale() {
--- a/src/libxls/locale.c
+++ b/src/libxls/locale.c
@@ -32,8 +32,24 @@
 #include <stdlib.h>
 #include "libxls/locale.h"
 
-xls_locale_t xls_createlocale() {
 #if defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
+#ifdef __GNUC__
+  // g++ 10 indicates Rtools42 and UCRT
+  #if __GNUC__ < 10
+    #define WINDOWS_BEFORE_RTOOLS_42
+  #endif
+#endif
+#endif
+
+#ifdef WINDOWS_BEFORE_RTOOLS_42
+static char* old_locale;
+#endif
+
+xls_locale_t xls_createlocale() {
+#if defined(WINDOWS_BEFORE_RTOOLS_42)
+    old_locale = setlocale(LC_CTYPE, ".65001");
+    return NULL;
+#elif defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
     return _create_locale(LC_CTYPE, ".65001");
 #else
     return newlocale(LC_CTYPE_MASK, "C.UTF-8", NULL);
@@ -41,6 +57,11 @@ xls_locale_t xls_createlocale() {
 }
 
 void xls_freelocale(xls_locale_t locale) {
+#if defined(WINDOWS_BEFORE_RTOOLS_42)
+    setlocale(LC_ALL, old_locale);
+    return;
+#endif
+
     if (!locale)
         return;
 #if defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
@@ -51,7 +72,9 @@ void xls_freelocale(xls_locale_t locale) {
 }
 
 size_t xls_wcstombs_l(char *restrict s, const wchar_t *restrict pwcs, size_t n, xls_locale_t loc) {
-#if defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
+#if defined(WINDOWS_BEFORE_RTOOLS_42)
+    return wcstombs(s, pwcs, n);
+#elif defined(_WIN32) || defined(WIN32) || defined(_WIN64) || defined(WIN64) || defined(WINDOWS)
     return _wcstombs_l(s, pwcs, n, loc);
 #elif defined(HAVE_WCSTOMBS_L)
     return wcstombs_l(s, pwcs, n, loc);
--- a/src/libxls/locale.c
+++ b/src/libxls/locale.c
@@ -60,7 +60,7 @@ xls_locale_t xls_createlocale() {
 
 void xls_freelocale(xls_locale_t locale) {
 #if defined(WINDOWS_BEFORE_RTOOLS_42)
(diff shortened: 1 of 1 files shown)


jennybc:
This looks good on my Windows VM, with display language set to German, with both R-devel and R 4.1.2.

Also looks good here on CI. This is a real result, because the first commit, where I simply restored the libxls v1.6.2 version of `locale.c`, failed on Windows 3.6.

Also looks good on win-builder:

`check_win_devel()` (R version 4.2.0 alpha (2022-03-25 r81986 ucrt)):
https://win-builder.r-project.org/KZM9PMD5m93Z/

`check_win_release()` (R version 4.1.3 (2022-03-10)):
https://win-builder.r-project.org/C4E0fQ6HVWg4/

`check_win_oldrelease()` (R version 4.0.5 (2021-03-31)):
https://win-builder.r-project.org/q6HrO921MN0g/

src/libxls/locale.c
```diff
@@ -32,29 +32,34 @@
...
-#if defined(__MINGW32__)
-  static char* old_locale;
+#ifdef WINDOWS_BEFORE_RTOOLS_42
+    static char* old_locale;
 #endif
 
 xls_locale_t xls_createlocale() {
-#if defined(__MINGW32__)
+#if defined(WINDOWS_BEFORE_RTOOLS_42)
     old_locale = setlocale(LC_CTYPE, ".65001");
```
jennybc, line 52: Note that this addresses `LC_CTYPE`.

src/libxls/locale.c
```diff
@@ -32,29 +32,34 @@
...
-    return NULL;
 #else
     return newlocale(LC_CTYPE_MASK, "C.UTF-8", NULL);
 #endif
 }
 
 void xls_freelocale(xls_locale_t locale) {
-#if defined(__MINGW32__)
+#if defined(WINDOWS_BEFORE_RTOOLS_42)
     setlocale(LC_ALL, old_locale);
```
jennybc, line 63: But this addresses `LC_ALL`.

This was already true of the patch, i.e. not something I've introduced. Does this seem wrong to you @gaborcsardi? I find it odd.

jennybc: I poked around this a bit more, inspecting `Sys.getlocale()` before and after a `read_xls()` and I conclude this should be `LC_CTYPE`. In fact, this may have been the only thing that absolutely needed to change, but I think other changes in this PR are also desirable. We really do want to distinguish R+Windows+UCRT from that which came before.

jennybc:
Re-checked again locally and all is still well.

Resubmitted to win-builder and all is still well:

Also looks good on win-builder:

`check_win_devel()` (R version 4.2.0 alpha (2022-03-26 r81993 ucrt)):
https://win-builder.r-project.org/F5u5OF3gE67y/

`check_win_release()` (R version 4.1.3 (2022-03-10)):
https://win-builder.r-project.org/PRPnU8eJ9gGF/

`check_win_oldrelease()` (R version 4.0.5 (2021-03-31)):
https://win-builder.r-project.org/gg6I1tI6211m/

src/libxls/locale.c
gaborcsardi: Yeah, I don't see any reason why it should not be `LC_CTYPE`.

jennybc:
And again.

Re-checked locally and all is still well. Also true for GHA.

Resubmitted to win-builder and all is still well:

`check_win_devel()` (R version 4.2.0 alpha (2022-03-26 r81993 ucrt)):
https://win-builder.r-project.org/M6LG1IpbzU1u/

`check_win_release()` (R version 4.1.3 (2022-03-10)):
tbd

`check_win_oldrelease()` (R version 4.0.5 (2021-03-31)):
tbd

gaborcsardi:
AFAICT we need at least Windows 8 to support `_create_locale()` and we also need support from MinGW.

What exactly is needed from MinGW is unclear for me even after several hours of investigation.

What seems clear is that since MinGW 7.0.0 `_create_locale()` is available on all Windows versions, and on older OSes it will be a stub, that does not return anything: https://github.com/mirror/mingw-w64/commit/87558c151f0a447d8f00af1a7d3a88a273aa8c23

In principle, it would be good to fix this upstream in libxls, so we don't need a custom patch. But that seems hard without spending even more time on investigating which option to use when.

As for the changes is this PR, the potential issue I see is that we decide between the option at compile time, but this is (also) a runtime issue. E.g. If the we build an R-devel binary package on Windows 10 or Windows server 2022 (like cran), then I am not sure if the MinGW stub will work well enough for us on Windows 7, which is still supported by R 4.2.x. But I can give this a try.

In theory the situation is similar for R 4.1.x and R 4.0.x. If we build the binary package on newer Windows, that might not work on older Windows. CRAN builds these binaries on Windows Server 8, so that is fine, but RSPM (?) and R-universe etc. use newer Windows, so it can happen there.

So I suggest we try what happens in these two cases, and if they work then we can go ahead with this PR. I have both a Vista and a Windows 7 VM, so I can try this in a couple of minutes.

gaborcsardi:
Yes, so if I build the PR on Windows 10, then I get a binary that does not work on Vista (and probably Windows 7).

gaborcsardi:
What is important, however, is that the PR currently does not build on R 4.1.x and 4.0.x on Windows 7 and Windows Vista, because these don't have `_create_locale()`. 

So if we keep the current approach, then we need to draw the line between R 4.2 and R 4.1.

jennybc:
As for this

> What I also don't understand, is that why the setlocale() solution that we had before (apparently) does not work on UCRT Windows

I think that might have been the `LC_CTYPE` / `LC_ALL` issue.

jennybc:
> What is important, however, is that the PR currently does not build on R 4.1.x and 4.0.x on Windows 7 and Windows Vista, because these don't have `_create_locale()`.

> So if we keep the current approach, then we need to draw the line between R 4.2 and R 4.1.

OK I could go back to that.

One last argument for leaving things as they are: it feels like those on these legacy systems will soon be feeling strong pressure from the whole R ecosystem to either upgrade Windows or to explicitly install UCRT themselves on older Windows, yeah? Is there really going to be a noticeable population of R users on Vista?

I do have a strong preference for having as few people as possible experiencing libxls, via readxl, with this patch we've added.

gaborcsardi:
>  Is there really going to be a noticeable population of R users on Vista?

Well CRAN is still building R-4.1 and R-4.0 binaries on a Vista (equivalent) machine.

So if you don't change it back, the package does not compile on CRAN's R 4.1 and R 4.0.

If you wait for R 4.2 to be out, then R 4.1 will be oldrel, and then they'll not care.

gaborcsardi:
> I think that might have been the LC_CTYPE / LC_ALL issue.

Ah, indeed, just confirmed this. So we could always use `setlocale()` on msvcrt builds. Is there anything against that?

jennybc:
Relevant:
https://stackoverflow.com/q/41844644

> ... MinGW-w64 never offered _create_locale/_free_locale (although they're defined in the headers, they don't exist in msvcrt.dll, which MinGW-w64 links against).

gaborcsardi:
Yes, it _might_ apply to some very old versions of MinGW. 

It definitely applies to Windows versions before Windows 8. But of course recent versions on MinGW provide a (not fully functional) replacement, so the situation is even more muddled.

jennybc:
OK, then, yeah this is why I wanted to note that here. And the murkiness re: how broadly it applies is also good to record.

gaborcsardi:
Now that I thought about this with a clear(er) head, I think the right thing to do is simple:

- use the system-provided `_create_locale()` on UCRT and/or ARM, because they are always available and work well.
- runtime check for `_create_locale()` on other systems, and provide a _proper_ fill-in if it is not available., including a fill-in for `_wcstombs_l()` as well. (Proper, in that it does not return `NULL`, so we can restore the locale easier.) MinGW's stub is not good enough, so we should completely ignore it.

PR #718, opened 2022-12-07
Fix warnings for CRAN

jeroen:
CRAN wants to use `snprintf()` instead of `sprintf()`. It is easy in this case, we can simply pass the sizes of the buffers that are created earlier in the function.


jeroen:
The easiest way is trying to get the patch into libxls itself, so that it is no longer your problem.

jennybc:
Hmmm, I can push into your PR but I guess all hell breaks loose if I force push (I was trying to rebase this onto `main`, among other things). Thanks for this PR. I will replicate it, using my new patch file strategy.

PR #722, opened 2023-02-08
Remove the C++11 declaration

jennybc:


DESCRIPTION
```diff
@@ -51,4 +51,3 @@ Encoding: UTF-8
 Note: libxls v1.6.2 (patched) 45abe77
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
-SystemRequirements: C++11
```
gaborcsardi, line 54: Since there is a `configure.win` file now, you need to add 
```
Biarch: true
```
here.

tools/configure.R
```diff
@@ -0,0 +1,13 @@
...
+
+  txt <- readLines(makevars)
+  txt_win <- readLines(makevars_win)
+
+  txt <- c("CXX_STD = CXX11", txt)
+  txt_win <- c("CXX_STD = CXX11", txt_win)
+
+  cat(txt, file = makevars, sep = "\n")
+  cat(txt_win, file = makevars_win, sep = "\n")
+}
```
gaborcsardi, line 13: One potential hiccup with this approach is that it modifies the files in place, so if you do multiple installs from the same directory (as opposed to package tarball), then those lines will be added multiple times.

Or, if you first install from the directory on R 3.5.x and then a newer version of R, then the newer version will also use C++11.

This is probably fine for this particular case, though, because not many people are doing active development on R 3.5.x.

gaborcsardi (approved):
Looks good! You only need to add a `Biarch: true` entry to `DESCRIPTION`.

It installs fine on Windows R 3.5.3, and also on CentOS 7 and R 3.5.3.

tools/configure.R
jennybc: I modified the code so that we don't add `CXX_STD` if something is already there. Then I noted the "multiple installs" concern in a comment.

jennybc:
@gaborcsardi As discussed on Slack, we're going to close this unmerged and use a simpler approach. But in the name of finishing the example, will you take one last look?

gaborcsardi:
Yep, all is good, I think.

jennybc:
We've decided the complexity of this is not worth it, since R 3.5 falls off the end of the tidyverse support policy within a couple of months.

PR #738, opened 2023-10-06
Fix link to tidyverse design principles

gl-eb:
Currently the link leads to [principles.tidyverse.org](https://principles.tidyverse.org/names-attribute.html), which is no longer active. The tidy design principles guide is now found at [design.tidyverse.org](https://design.tidyverse.org/names.html).


jennybc:
Thanks!

Issue #710, opened 2022-09-20
read_excel() crashes R session


ja-ortiz-uniandes:
Before I start thanks to the `readxl` team, your product is awesome!
# Problem:
I have a data set in Excel I'm try to import into R. The file opens without issue. The data starts on the 6th line and I only wanted to import a part of the file so I did:
```
library(readxl)
read_excel("myfile.xslx", range = "A7:P100")
```
I immediately get this screen:
![R crash](https://imgur.com/VXYwOWn.png)

# Reprex
After [downloading the file](https://filedropper.com/d/s/0B3oelp1L1IXYqqUiRWaZDDqMvGiKP) simply do:
```
readxl::read_excel(file, range = "A7:P100")
```
where `file` is the path.
# Expected behavior
1. No session crash
2. Get the data in the A7:P100 range imported into R

ja-ortiz-uniandes:
I think someone should take a look at this as it could be a critical error.

jennybc:
Please provide the problematic file as described below.

## How to provide a readxl reprex

We're in a much better position to address your issue if you can provide a reprex (**repr**oducible **ex**ample). Provide as much of this as you can:

  * An actual xls or xlsx file. Pick one:
    - Your personal xls or xlsx: try to strip it down to the minimal size and complexity to demonstrate your point. And, obviously, remove any sensitive data.
    - A publicly available xls or xlsx: provide URL and the code you used to download.
  * A small bit of R code that uses readxl on the provided xls or xlsx file and demonstrates your point.
    - Consider using the [reprex package](https://github.com/jennybc/reprex#readme) to prepare this. In addition to nice formatting, this ensures your reprex is self-contained.
  * Any details about your environment that seem clearly relevant, such as operating system.
        `reprex(..., si = TRUE)`
    will append a standard summary, folded neatly away, at the bottom of your reprex.

How to provide your own xls/xlsx file? In order of preference:

  1. [Attach the file directly to your issue](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/). Instructions are always at the bottom of the issue or comment box. `.xlsx` is a supported file type. You'll need to zip or gzip `.xls` so it appears as `.zip` or `.gz`.
  1. Share via DropBox or Google Drive and provide the link in your issue.
  1. Explain you absolutely cannot provide a relevant file via github.com and offer to provide privately.
  1. Don't share a file and realize you're hoping for, e.g., a bug fix with no concrete example to go on.

[...]

PR #742, opened 2024-03-11
Use negative or infinite sheet number to select sheet counting from the last

DanChaltiel:
Hi,

It is quite a natural interface to select the last sheet using a negative integer or infinite, one that I've seen in many packages.

Moreover, as noticed in https://github.com/tidyverse/readxl/issues/710, using `sheet=Inf` plainly makes the R session crash.

This little PR circumvolves this problem. 

Unfortunately, I could not manage to make the compiler work so I only tested that using `assignInNamespace()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # readxl (development version)
 
+# readxl 1.5.0
+
+* Sheet number in `read_excel()`, `read_xls()`, and `read_xlsx()` can now be negative or infinite to select the sheet counting from the last one (#742, @DanChaltiel).
+
 # readxl 1.4.3
 
 This release contains no user-facing changes.
--- a/R/read_excel.R
+++ b/R/read_excel.R
@@ -237,9 +237,10 @@ standardise_sheet <- function(sheet, range, sheet_names) {
 
   if (is.numeric(sheet)) {
     if (sheet < 1) {
-      stop("`sheet` must be positive", call. = FALSE)
+      length(sheet_names) + sheet
+    } else {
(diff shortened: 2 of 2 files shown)


jennybc:
I am not keen to support this behaviour around negative or infinite sheet number. I do know where you're coming from, but negative selections also often mean "not this one" in base R and the tidyverse. So I think it's OK for someone with this need to do this instead:

``` r
library(readxl)

path <- readxl_example("datasets.xlsx")
sheets <- excel_sheets(path)

path |> 
  read_excel(sheet = sheets[length(sheets)])
#> # A tibble: 1,000 × 5
#>      lat  long depth   mag stations
#>    <dbl> <dbl> <dbl> <dbl>    <dbl>
#>  1 -20.4  182.   562   4.8       41
#>  2 -20.6  181.   650   4.2       15
#>  3 -26    184.    42   5.4       43
#>  4 -18.0  182.   626   4.1       19
#>  5 -20.4  182.   649   4         11
#>  6 -19.7  184.   195   4         12
#>  7 -11.7  166.    82   4.8       43
#>  8 -28.1  182.   194   4.4       15
#>  9 -28.7  182.   211   4.7       35
#> 10 -17.5  180.   622   4.3       19
#> # ℹ 990 more rows
```

<sup>Created on 2026-09-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

However, I am glad you raised the point about bad behaviour when `sheet = Inf`! That inspired me to to add more input checks for `sheet` in #801.

Issue #642, opened 2021-04-23
Ditching iris


mine-cetinkaya-rundel:
The first sheet in `datasets.xlsx` is `iris`. Would be good to replace that with another dataset, or just remove it and use the remaining sheets.

PR #751, opened 2024-08-15
dev day, remove iris from README (RMD + MD). #642

drorberel:
issue #642
partial change. relevant only to README file.
See another PR for other related tasks.
These independent PRS should not conflict with each other.

Jointly with @fontikar


jennybc:
Looks like we already got these changes via #752.

PR #752, opened 2024-08-15
Don't mention iris dataset in documentation 

fontikar:
Addresses #642 

Problem:  `iris` dataset is stored in `dataset.xlsx` and `dataset.xls`. `iris` is used in various examples to illustrate the functionality of `readxl`, the goal is to remove this `iris` or replace it with another nicer dataset e.g. penguins 🐧 

After chats with Garrick and Emil, they suggested the smallest scoped solution is to remove any public facing documentation is uses `iris` and ignore the test files that uses iris. 

@drorberel and I decided to remove the iris spreadsheet and update the README files, vignettes and any examples in .R that refer to `iris`. 

Hope that we didn't cause any downstream breakages! 😄


R/read_excel.R
```diff
@@ -57,13 +57,13 @@ NULL
...
 #' # Skip rows and use default column names
-#' read_excel(datasets, skip = 148, col_names = FALSE)
+#' read_excel(datasets, skip = 10, col_names = FALSE)
 #'
 #' # Recycle a single column type
 #' read_excel(datasets, col_types = "text")
 #'
 #' # Specify some col_types and guess others
-#' read_excel(datasets, col_types = c("text", "guess", "numeric", "guess", "guess"))
+#' read_excel(datasets, col_types = rep("guess", 11))
```
jennybc, line 66: This change directly contradicts the comment just above it. Can you target a different worksheet and hold on to the intent of the example?

jennybc (commented):
Looking good! There's one example I hope we can tweak.

R/read_excel.R
fontikar: So true! Thanks for pointing this out 😊 I've updated this now with a better example!

jennybc:
Thanks!

PR #757, opened 2024-08-23
Avoid scary implicit conversion from `sexp` to `bool`

DavisVaughan:
cpp11 allows some somewhat scary implicit conversions between `sexp` and the 3 types `bool`, `size_t`, and `double`. These are unchecked conversions, and we'd like to remove them:
https://github.com/r-lib/cpp11/pull/390/

The recommended way is to use `cpp11::as_cpp<bool>()`, which checks the type and length before performing the conversion.

I'm not going to remove these in this version of cpp11, but we will in the next

### DIFF
--- a/src/zip.cpp
+++ b/src/zip.cpp
@@ -2,8 +2,10 @@
 
 #include "rapidxml/rapidxml_print.h"
 
+#include "cpp11/as.hpp"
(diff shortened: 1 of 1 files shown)


jennybc (approved):
Thanks!

PR #760, opened 2024-10-25
Fix compilation on NetBSD

gaborcsardi:
NetBSD has wcstombs_l(), but it does not have
uselocale().


jennybc:
This is a generated file, yeah? Do you have any thoughts about that aspect?

gaborcsardi:
Oh, sorry, I didn't notice. How is it generated?

jennybc:
> How is it generated?

With `configure`, in libxls:

https://github.com/libxls/libxls?tab=readme-ov-file#installation

But we have certainly been known to layer manual tweaks on top of that result.

PR #763, opened 2025-01-02
fix link

### LINKED ISSUE #755: vignette ‘readxl-workflows’ not found
The documentation for `excel_sheets()` reads:

> The list of sheet names is especially useful when you want to iterate over all of the sheets in a workbook. The `vignette("readxl-workflows")` article provides several worked examples of this, showing how to combine readxl with other packages in the tidyverse, such as purrr, or with base R functions like lapply().

However, this vignette isn't exported by `{readxl}`:

```
Vignettes in package ‘readxl’:

cell-and-column-types                           Cell and Column Types (source, html)
sheet-geometry                                  Sheet Geometry (source, html)
```

I can see it's on the `{pkgdown}` website, but it's not a vignette available locally, which threw me off trying to open it using `vignette()` like the docs suggest!

olivroy:
Fix #755 + sneak in a small r-devel anchored link issue

### DIFF
--- a/R/cell-specification.R
+++ b/R/cell-specification.R
@@ -7,7 +7,7 @@
 #' using helper functions. You can find more examples at the
 #'  [sheet geometry](https://readxl.tidyverse.org/articles/sheet-geometry.html#range)
 #'  vignette. In all cases, cell range processing is handled by the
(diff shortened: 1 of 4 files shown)


jennybc:
Thanks!

PR #785, opened 2026-04-30
Avoid cpp11::warning / cpp11::stop variadic forwarding

### LINKED ISSUE #784: Segfault on MacOS
I am seeing the same segfault on r-universe as in the CI here: https://github.com/r-universe/tidyverse/actions/runs/24253882831/job/70947939190

jennybc:
Closes #784

readxl segfaults in various places:

- r-universe (segfault; reported in #784)
- CRAN: `r-devel-linux-x86_64-{debian,fedora}-clang` (segfault); `r-release-macos-x86_64` (segfault);
 https://cran.r-project.org/web/checks/check_results_readxl.html.
- Local: macOS arm64, Apple clang 21, with the "right" compilation flags.

Whether you see a segfault or some other failure mode (e.g. indefinitie hang) appear to depend on compilation flags. Some of these other presentations are seen:

- CRAN's additional checks on `M1Mac` ("Warning: elapsed-time limit of 1.5 hours reached for sub-process")
- readxl's own CI on macos-latest, e.g. https://github.com/tidyverse/readxl/actions/runs/25193092291/job/73867448241. I've had to just kill those jobs.
- Locally, with default compilation flags.

From bisection, it appears to be a regression triggered by upgrading cpp11 from 0.5.3 to 0.5.4 (released 2026-04-09 on CRAN).

Claude Code analysis: lldb shows the crash deep inside cpp11's `safe[Rf_warningcall]` machinery — specifically `std::tuple_leaf::get(this=0x0000ffff01000000)` while `apply()` is unpacking the tuple of forwarded references that the closure passed to `R_UnwindProtect` was built from. The leaf address is corrupted (the 0xffff... pattern looks PAC-failure-like, but the same crash hits Linux clang 21 with no PAC, so likely a generic codegen/lifetime issue rather than PAC).

Workaround taken in this PR: format warning and stop messages into a `std::string` ourselves and call `Rf_warningcall` / `Rf_errorcall` directly inside `cpp11::unwind_protect`. This keeps longjmp protection but avoids cpp11's variadic-template arg forwarding entirely.

Filed upstream at https://github.com/r-lib/cpp11/issues/491

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -6,6 +6,7 @@
 #include <cmath>
 #include <cstring>
 #include <sstream>
+#include <string>
 
 //May appear in cpp11
 template <typename T, typename N>
@@ -13,6 +14,37 @@ T new_vector(R_xlen_t size, N val) {
   T out(size);
   std::fill(out.begin(), out.end(), val);
   return out;
+}
+
+// Pre-formatted warning/error helpers that avoid cpp11's variadic-template
+// forwarding for warning()/stop(). With clang 21 + R 4.6 on Apple Silicon
+// (and Linux clang on CRAN), `cpp11::warning(fmt, args...)` segfaults inside
+// its closure/tuple forwarding to Rf_warningcall. Wrapping a direct call to
+// Rf_warningcall / Rf_errorcall in cpp11::unwind_protect keeps longjmp
+// protection without going through the broken path.
+inline void readxl_warning(const char* msg) {
+  cpp11::unwind_protect([&] {
+    Rf_warningcall(R_NilValue, "%s", msg);
+  });
+}
+
+inline void readxl_warning(const std::string& msg) {
+  readxl_warning(msg.c_str());
+}
+
+[[noreturn]] inline void readxl_stop(const char* msg) {
+  cpp11::unwind_protect([&] {
+    Rf_errorcall(R_NilValue, "%s", msg);
+  });
+  // Unreachable: Rf_errorcall longjmps and cpp11::unwind_protect rethrows it
+  // as a unwind_exception. The compiler can't see through that, so hint it
+  // explicitly to satisfy [[noreturn]] without leaving a real terminator
+  // (R CMD check flags `abort` / `exit` in compiled code).
+  __builtin_unreachable();
+}
+
+[[noreturn]] inline void readxl_stop(const std::string& msg) {
+  readxl_stop(msg.c_str());
 }
 
 // The date offset needed to align Excel dates with R's use of 1970-01-01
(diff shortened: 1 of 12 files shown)


src/utils.h
```diff
@@ -15,6 +16,37 @@ T new_vector(R_xlen_t size, N val) {
...
 
+// Pre-formatted warning/error helpers that avoid cpp11's variadic-template
+// forwarding for warning()/stop(). With clang 21 + R 4.6 on Apple Silicon
+// (and Linux clang on CRAN), `cpp11::warning(fmt, args...)` segfaults inside
+// its closure/tuple forwarding to Rf_warningcall. Wrapping a direct call to
+// Rf_warningcall / Rf_errorcall in cpp11::unwind_protect keeps longjmp
+// protection without going through the broken path.
+inline void readxl_warning(const char* msg) {
+  cpp11::unwind_protect([&] {
+    Rf_warningcall(R_NilValue, "%s", msg);
```
DavisVaughan, line 27:
```suggestion
    Rf_warningcall(R_NilValue, msg);
```

is there a reason not to do this? something about string injection maybe? but readxl controls all of the paths, so it would not worry me to do this instead

src/utils.h
```diff
@@ -15,6 +16,37 @@ T new_vector(R_xlen_t size, N val) {
...
+  });
+}
+
+inline void readxl_warning(const std::string& msg) {
+  readxl_warning(msg.c_str());
+}
+
+[[noreturn]] inline void readxl_stop(const char* msg) {
+  cpp11::unwind_protect([&] {
+    Rf_errorcall(R_NilValue, "%s", msg);
```
DavisVaughan, line 37:
```suggestion
    Rf_errorcall(R_NilValue, msg);
```

same thought here

src/utils.h
```diff
@@ -15,6 +16,37 @@ T new_vector(R_xlen_t size, N val) {
...
+  });
+  // Unreachable: Rf_errorcall longjmps and cpp11::unwind_protect rethrows it
+  // as a unwind_exception. The compiler can't see through that, so hint it
+  // explicitly to satisfy [[noreturn]] without leaving a real terminator
+  // (R CMD check flags `abort` / `exit` in compiled code).
+  __builtin_unreachable();
```
DavisVaughan: that's almost certainly not going to fly on all platforms

we do this
https://github.com/r-lib/cpp11/blob/a9df50ef272e098ebb7b18e504fb50f50ded5a06/inst/include/cpp11/protect.hpp#L177

DavisVaughan (approved):
Looks fine to me with a few minor potential nits

src/utils.h
jennybc: Well, we potentially inline the content of the cell, e.g.:

```
#> Warning: Expecting logical in A8 / R8C1: got 'cabbage'
```

and *that* could have `%` or who knows what. So it seems safest to keep the indirection.

PR #798, opened 2026-09-09
Reject NA sheet and integer-overflow limits in read_excel()

sims1253:
### sheet = NA crashes with an internal error

standardise_sheet() checked the sheet number with `if (sheet < 1)`. A numeric NA made the comparison NA and R failed before reaching the intended message:

```r
read_excel(path, sheet = NA_real_)
#> Error: missing value where TRUE/FALSE needed
```

The positivity check now rejects NA first and keeps the same error.

### Large limits overflow to NA and drop the read cap

standardise_limits() maps NA to -1, the "unspecified" sentinel, and did so before converting to integer. A limit above .Machine$integer.max overflowed to NA inside as.integer() and passed through as unspecified, so the cap disappeared. A 150-row sheet read with n_max = 1e10 returned all 150 rows:

```r
read_excel(path, n_max = 1e10)
#> Warning: NAs introduced by coercion to integer range
#> # A tibble: 150 x 5
```

The conversion now runs first, entries that were NA only after the conversion trigger an error naming the offending limit, and only originally-NA entries become -1.
With n_max = 1e10 the error is "`max_row` limit is too large to be represented as an integer".

Found during my most recent [ry](https://github.com/sims1253/ry) audit.


jennybc:
Thanks! This produces better error messages. I don't think the integer-overflow issue is really a problem in actual practice, since xlsx and xls files can only have 2^20 and 2^16 rows (way below `.Machine$integer.max`). But it's fine to guard against it.

PR #799, opened 2026-09-21
Update assert handling

jennybc:
Backporting changes CRAN made recently.

I received an email about this CRAN-authored release on 2026-09-16 alluding to an email that was allegedly sent on 2026-08-26. I can't find evidence of that email, hence I was not aware that readxl was under a deadline. This PR backports the substantive change and cleans it up a little.


jennybc:
The now-removed treatment of `assert` was introduced in 81443a1c4cb97511e2ec31cb77f902edf6cc18bd and I'd just never (re-)examined that.

The most relevant snippet from [Writing R Extensions](https://rstudio.github.io/r-manuals/r-exts/Creating-R-packages.html#writing-portable-packages):

> Under no circumstances should your compiled code ever call `abort` or `exit`: these terminate the user’s R process, quite possibly losing all unsaved work. One usage that could call `abort` is the `assert` macro in C or C++ functions, which should never be active in production code. The normal way to ensure that is to define the macro `NDEBUG`, and `R CMD INSTALL` does so as part of the compilation flags. Beware of including headers (including from other packages) which could undefine it, now or in future versions.

